PR #7, opened 2024-03-13
Detect failures to close socket in FreeRTOS.

hlef:
Unlike the API reference suggests, `FreeRTOS_closesocket` *can* actually fail if the event queue is full.  This is really bad, as it will make us leak the socket allocation. Before this commit, this failure is silently ignored, which may lead to the system silently running OOM.

This commit does not fix the leak, however it logs the failure to ensure that such issues are more easily diagnosed.


davidchisnall:
Thanks! I think that we probably want to reorder the creation path to allocate memory so that the socket create is first and we don’t lose it. Similarly, on the destroy path, if we do a `heap_can_free` first, then close the socket, then free things, if we fail to close the socket then we’re still in a good state. Don’t upgrade the lock for destruction until you’ve closed the socket.

hlef:
Noted. I can come back to this PR later to do that.

lib/tcpip/network_wrapper.cc
```diff
@@ -444,17 +456,19 @@ int network_socket_close(Timeout *t, SObj mallocCapability, SObj sealedSocket)
...
+		  // terminating the socket, ensure that the frees won't fail
+		  if (heap_can_free(mallocCapability, socket->socket) != 0 ||
+		      token_obj_can_destroy(
+		        mallocCapability, socket_key(), sealedSocket) != 0)
 		  {
-			  Debug::log("Failed to free socket");
-			  // Release the lock so that we don't leak it.
-			  socket->socketLock.unlock();
-			  return -EINVAL;
+			  return -EAGAIN; // TODO EINVAL?
```
davidchisnall, line 465: Should be `EINVAL` or `EPERM`: it's definitely not `EAGAIN`: retrying with the wrong thing won't make it right.

lib/tcpip/network_wrapper.cc
```diff
@@ -364,7 +364,13 @@ SObj network_socket_create_and_bind(Timeout       *timeout,
 	if (!c)
 	{
 		Debug::log("Failed to claim socket");
-		FreeRTOS_closesocket(socket);
+		if (FreeRTOS_closesocket(socket) != 1)
+		{
+			// Nothing we can do to recover here. Make sure we log the failure -
+			// we will leak the socket allocation.
```
davidchisnall, line 370: Can this fail for any reason other than the queue being full?  If not, then we may need to yield and retry here.  We may also want to return a closed-but-allocated socket here.  This is partly why I wanted to do all of the allocations first: we can always free things, we may not be able to close the socket, so if we allocate the memory first then we can always return a socket that immediately returns `-ENOTCONNECTED` on send / receive and triggers the caller to try to close it again.

hlef: I implemented the yield and retry.

I would not do the "return a closed-but-allocated socket" part. This seems to require non-trivial and error-prone changes on the return path which probably aren't really worth it given that we now have the "yield and retry".

hlef: One thing that I found really tricky is the situation where the claim fails. If we return that socket to the user, all sorts of weird things can happen if the socket gets freed under the hood... It also requires us to change the close path to consider that `heap_can_free` can fail if we didn't claim successfully, which in turn calls for more changes...

The second option that you mention sounds good. I will try to implement it. I suggest that we leave the "return a closed socket" thing for later.

hlef: Alright, implemented it.

I tested it by simulating a failure in `FreeRTOS_closesocket`:
```
diff --git a/source/FreeRTOS_Sockets.c b/source/FreeRTOS_Sockets.c
index 6f495e6..c66a8c6 100644
--- a/source/FreeRTOS_Sockets.c
+++ b/source/FreeRTOS_Sockets.c
@@ -1983,6 +1983,8 @@ BaseType_t vSocketBind( FreeRTOS_Socket_t * pxSocket,
 } /* Tested */
 /*-----------------------------------------------------------*/
 
+static int counter = 0;
+
 /**
  * @brief Close a socket and free the allocated space. In case of a TCP socket:
  *        the connection will not be closed automatically. Subsequent messages
@@ -2015,6 +2017,11 @@ BaseType_t FreeRTOS_closesocket( Socket_t xSocket )
     }
     else
     {
+       if (counter != 2)
+       {
+               counter++;
+               return -1;
+       }
         #if ( ipconfigUSE_CALLBACKS == 1 )
             {
                 #if ( ipconfigUSE_TCP == 1 )
```

and then simulating a failure in `network_socket_create_and_bind`:
```
diff --git a/lib/tcpip/network_wrapper.cc b/lib/tcpip/network_wrapper.cc
index 7a85f89..e65468b 100644
--- a/lib/tcpip/network_wrapper.cc
+++ b/lib/tcpip/network_wrapper.cc
@@ -17,7 +17,7 @@
 #include <platform-ethernet.hh>
 #include <token.h>
 
-using Debug            = ConditionalDebug<false, "Network stack wrapper">;
+using Debug            = ConditionalDebug<true, "Network stack wrapper">;
 constexpr bool UseIPv6 = CHERIOT_RTOS_OPTION_IPv6;
 
 using CHERI::Capability;
@@ -431,7 +431,7 @@ SObj network_socket_create_and_bind(Timeout       *timeout,
        // network stack also keeps a claim to it.  We will drop this claim on
        // deallocation.
        Claim c{mallocCapability, socket};
-       if (!c)
+       if (c)
        {
                Debug::log("Failed to claim socket.");
                // Note that `close_socket_retry` can fail, in which case we
```

It would be nice to check that the IP thread runs in between but I'm not too sure on how to test that.

davidchisnall (approved):

PR #8, opened 2024-03-14
Various fixes to MQTT

hlef:
This PR ships many fixes to many more or less important issues in the recently pushed MQTT support.

Some of these fixes were proposed by David in an experimental branch - I cleaned them up and included them in relevant commits.

### DIFF
--- a/examples/04.MQTT/mqtt.cc
+++ b/examples/04.MQTT/mqtt.cc
@@ -17,12 +17,14 @@ using CHERI::Capability;
 using Debug            = ConditionalDebug<true, "MQTT example">;
 constexpr bool UseIPv6 = CHERIOT_RTOS_OPTION_IPv6;
 
-// The client ID is used by the broker to identify the client. It should be
-// unique per client. Since this is hard-coded, this means we cannot launch
-// multiple concurrent instances of this example on a same test server.
-// TODO it would be nice to generate automatically.
-constexpr const char  *clientID       = "anoncheri42";
-constexpr const size_t clientIDLength = sizeof(clientID);
+/// Maximum permitted MQTT client identifier length (from the MQTT
+/// specification)
+constexpr size_t MQTTMaximumClientLength = 23;
+/// Prefix for MQTT client identifier
+constexpr std::string_view clientIDPrefix{"cheriotMQTT"};
+/// Space for the random client ID.
+std::array<char, MQTTMaximumClientLength> clientID;
+static_assert(clientIDPrefix.size() < clientID.size());
 
 // MQTT network buffer sizes
 constexpr const size_t networkBufferSize    = 1024;
@@ -114,6 +116,13 @@ void __cheri_compartment("mqtt_example") example()
 	auto heapAtStart =
 	  heap_quota_remaining(STATIC_SEALED_VALUE(mqttTestMalloc));
 
+	Debug::log("Generating client ID...");
+	// Prefix with something recognizable, for convenience.
+	memcpy(clientID.data(), clientIDPrefix.data(), clientIDPrefix.size());
+	// Suffix with random character chain.
+	mqtt_generate_client_id(clientID.data() + clientIDPrefix.size(),
+	                        clientID.size() - clientIDPrefix.size());
+
 	Debug::log("Connecting to MQTT broker...");
 
 	SObj handle = mqtt_connect(&t,
@@ -126,8 +135,8 @@ void __cheri_compartment("mqtt_example") example()
 	                           networkBufferSize,
 	                           incomingPublishCount,
 	                           outgoingPublishCount,
-	                           clientID,
-	                           clientIDLength);
+	                           clientID.data(),
+	                           clientID.size());
 
 	if (!Capability{handle}.is_valid())
 	{
--- a/examples/04.MQTT/mqtt.cc
+++ b/examples/04.MQTT/mqtt.cc
@@ -227,7 +227,7 @@ void __cheri_compartment("mqtt_example") example()
 	Debug::log("Disconnecting from the broker.", testTopic);
 
 	t   = Timeout{MS_TO_TICKS(5000)};
-	ret = mqtt_disconnect(&t, handle);
+	ret = mqtt_disconnect(&t, STATIC_SEALED_VALUE(mqttTestMalloc), handle);
 
 	if (ret < 0)
 	{
--- a/examples/04.MQTT/mqtt.cc
+++ b/examples/04.MQTT/mqtt.cc
@@ -40,8 +40,8 @@ DECLARE_AND_DEFINE_CONNECTION_CAPABILITY(MosquittoOrgMQTT,
 
 DECLARE_AND_DEFINE_ALLOCATOR_CAPABILITY(mqttTestMalloc, 32 * 1024);
 
-const char *testTopic   = "cherries";
-const int   testPayload = 42;
+constexpr std::string_view testTopic{"cherries"};
+constexpr std::string_view testPayload{"MORELLO"};
 
 static int ackReceived     = 0;
 static int publishReceived = 0;
@@ -151,8 +151,8 @@ void __cheri_compartment("mqtt_example") example()
 	ret = mqtt_subscribe(&t,
 	                     handle,
 	                     1, // QoS 1 = delivered at least once
-	                     testTopic,
-	                     sizeof(testTopic));
+	                     testTopic.data(),
+	                     testTopic.size());
 
 	if (ret < 0)
 	{
@@ -180,10 +180,10 @@ void __cheri_compartment("mqtt_example") example()
 	ret = mqtt_publish(&t,
 	                   handle,
 	                   1, // QoS 1 = delivered at least once
-	                   testTopic,
-	                   sizeof(testTopic),
-	                   static_cast<const void *>(&testPayload),
-	                   sizeof(testPayload));
+	                   testTopic.data(),
+	                   testTopic.size(),
+	                   static_cast<const void *>(testPayload.data()),
+	                   testPayload.size());
 
 	if (ret < 0)
 	{
@@ -212,8 +212,8 @@ void __cheri_compartment("mqtt_example") example()
 	ret = mqtt_unsubscribe(&t,
 	                       handle,
 	                       1, // QoS 1 = delivered at least once
-	                       testTopic,
-	                       sizeof(testTopic));
+	                       testTopic.data(),
+	                       testTopic.size());
 
 	if (ret < 0)
 	{
@@ -233,7 +233,7 @@ void __cheri_compartment("mqtt_example") example()
 		}
 	}
 
-	Debug::log("Disconnecting from the broker.", testTopic);
+	Debug::log("Disconnecting from the broker.");
 
 	t   = Timeout{MS_TO_TICKS(5000)};
 	ret = mqtt_disconnect(&t, STATIC_SEALED_VALUE(mqttTestMalloc), handle);
--- a/include/mqtt.h
+++ b/include/mqtt.h
@@ -123,7 +123,9 @@ int __cheri_compartment("MQTT")
  * Publish on a given MQTT connection.
  *
  * The parameters `topic` and `payload` (and related lengths `topicLength` and
- * `payloadLength`) indicate the topic and payload.
+ * `payloadLength`) indicate the topic and payload. `topic` must be a valid
+ * MQTT 3.1.1 topic (i.e., it must not include the zero terminator and must be
+ * at least one-character long).
  *
  * `qos` indicates the level of QoS (0, 1, or 2).
  *
@@ -165,7 +167,9 @@ int __cheri_compartment("MQTT") mqtt_publish(Timeout    *t,
 /**
  * Subscribe on a given MQTT connection.
  *
- * The parameters `filter` indicates on which topic to subscribe.
+ * The parameter `filter` (of length `filterLength`) indicates on which topic
+ * to subscribe. `filter` must be a valid MQTT 3.1.1 filter (i.e., it must not
+ * include the zero terminator and must be at least one-character long).
  *
  * `qos` indicates the level of QoS (0, 1, or 2).
  *
@@ -204,7 +208,9 @@ int __cheri_compartment("MQTT") mqtt_subscribe(Timeout    *t,
 /**
  * Unsubscribe on a given MQTT connection.
  *
- * The parameters `filter` indicates on which topic to unsubscribe.
+ * The parameter `filter` (of length `filterLength`) indicates on which topic
+ * to unsubscribe. `filter` must be a valid MQTT 3.1.1 filter (i.e., it must
+ * not include the zero terminator and must be at least one-character long).
  *
  * `qos` indicates the level of QoS (0, 1, or 2).
  *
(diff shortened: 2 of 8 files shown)


examples/04.MQTT/mqtt.cc
```diff
@@ -114,6 +112,9 @@ void __cheri_compartment("mqtt_example") example()
 	auto heapAtStart =
 	  heap_quota_remaining(STATIC_SEALED_VALUE(mqttTestMalloc));
 
+	Debug::log("Generating client ID...");
+	mqtt_generate_client_id(clientID, clientIDLength);
```
davidchisnall, line 116: It would be nicer here to have a prefix and a random suffix. The same API can be used, just pass the tail of the string rather than the whole thing.

include/mqtt.h
```diff
@@ -87,13 +92,16 @@ SObj __cheri_compartment("MQTT")
                size_t                      incomingPublishCount,
                size_t                      outgoingPublishCount,
                const char                 *clientID,
-               size_t                      clientIDLength);
+               size_t                      clientIDLength,
+               bool                        newSession = true);
```
davidchisnall, line 96: This is not valid C. All public headers should compile as C and C++. We have an `__if_CXX` macro for things like this.

lib/mqtt/mqtt.cc
```diff
@@ -108,19 +107,12 @@ namespace
...
+	uint64_t rand()
 	{
-		// If the link fails with -ENOTCONN as part of `transport_recv`
-		// or `transport_send`, the `tlsHandle` is invalidated.
-		if (!Capability{tlsHandle}.is_valid())
-		{
-			return true;
-		}
-		return false;
+		EntropySource rng;
```
davidchisnall, line 114: This should be static.

lib/mqtt/mqtt.cc
```diff
@@ -629,17 +625,16 @@ SObj mqtt_connect(Timeout                    *t,
 		return nullptr;
 	}
 
-	// Since we requested a clean session, sessionPresent should be false
-	Debug::Assert(sessionPresent == false,
+	Debug::Assert(sessionPresent == !newSession,
 	              "Successfully connected, but a previous session was "
-	              "unexpectedly present.");
+	              "unexpectedly present or absent.");
```
davidchisnall, line 630: This is confusing, maybe make it conditional on new session to pass present or absent as a for,at argument?

lib/mqtt/mqtt.cc
```diff
@@ -954,3 +943,42 @@ int mqtt_run(Timeout *t, SObj mqttHandle)
...
+	 * Note from the MQTT 3.1.1 spec:
+	 * The Server MUST allow ClientIds which are between 1 and 23 UTF-8 encoded
+	 * bytes in length, and that contain only the characters
+	 * "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+	 *
+	 * Note from us:
+	 * UTF-8 encoding of 0-9, a-z, A-Z, is 1 Byte per character, so we should be
+	 * able to do up to a length of 22 characters + zero byte.
+	 */
+	const char characters[] =
```
davidchisnall, line 959: If you make this a std::string_view, you can use .size() instead of sizeof calculations.

lib/mqtt/mqtt.cc
hlef: That makes sense. Actually I copied this from the SNTP code, which probably also needs the static keyword.

lib/mqtt/mqtt.cc
hlef: For some reason I'm hitting an unrelated bug with std::string_view, so I would rather leave this for future works.

examples/04.MQTT/mqtt.cc
```diff
@@ -17,12 +17,12 @@ using CHERI::Capability;
...
-// multiple concurrent instances of this example on a same test server.
-// TODO it would be nice to generate automatically.
-constexpr const char  *clientID       = "anoncheri42";
-constexpr const size_t clientIDLength = sizeof(clientID);
+constexpr const int clientIDLength = 23;
+char                clientID[clientIDLength];
+
+constexpr const char clientIDPrefix[]     = "cheriotMQTT";
+constexpr const int  clientIDPrefixLength = sizeof(clientIDPrefix) - 1 /* \0 */;
+static_assert(clientIDPrefixLength < clientIDLength);
```
davidchisnall, line 25: It's better style to make this correct by construction, rather than to add asserts that check that you got it right.  For example:

```c++
/// Prefix for MQTT client identifier
constexpr const char clientIDPrefix[]     = "cheriotMQTT";
/// Length of client identifier 
constexpr const int  clientIDPrefixLength = sizeof(clientIDPrefix) - 1;
/// Length of the client ID including the suffix
constexpr const int clientIDLength = clientIDPrefixLength + 12;
/// Space for the random client ID.
char                clientID[clientIDLength];
```

This can be further simplified as:

```c++
/// Prefix for MQTT client identifier
constexpr std::string_view clientIDPrefix{"cheriotMQTT"};
/// Length of the client ID including the suffix
constexpr const int clientIDLength = clientIDPrefix.size() + 12;
/// Space for the random client ID.
std::array<char, clientIDLength> clientID;
```

EDIT: It looks as if the 23 number came from the maximum client ID length permitted by MQTT.  In that case, that should be a symbolic constant and then we possibly do need the static assert, but then the code looks like this:

```c++
/// Maximum permitted MQTT client identifier length (from the MQTT specification)
constexpr size_t MQTTMaximumClientLength = 23;
/// Prefix for MQTT client identifier
constexpr std::string_view ClientIDPrefix{"cheriotMQTT"};
/// Space for the random client ID.
std::array<char, MQTTMaximumClientLength> clientID;
static_assert(ClientIDPrefix.size() < clientID.size);
```

examples/04.MQTT/mqtt.cc
```diff
@@ -38,8 +38,10 @@ DECLARE_AND_DEFINE_CONNECTION_CAPABILITY(MosquittoOrgMQTT,
 
 DECLARE_AND_DEFINE_ALLOCATOR_CAPABILITY(mqttTestMalloc, 32 * 1024);
 
-const char *testTopic   = "cherries";
-const int   testPayload = 42;
+const char   testTopic[]     = "cherries";
+const size_t testTopicSize   = sizeof(testTopic) - 1;
+const int    testPayload     = 42;
+const size_t testPayloadSize = sizeof(testPayload) - 1;
```
davidchisnall, line 44: You can avoid all of the `-1` calculations here using `constexpr std::string_view` and `.size()`.

examples/04.MQTT/mqtt.cc
```diff
@@ -114,6 +116,13 @@ void __cheri_compartment("mqtt_example") example()
 	auto heapAtStart =
 	  heap_quota_remaining(STATIC_SEALED_VALUE(mqttTestMalloc));
 
+	Debug::log("Generating client ID...");
+	// Prefix with something recognizable, for convenience.
+	memcpy(clientID, clientIDPrefix, clientIDPrefixLength);
```
davidchisnall, line 121: Please don't use `memcpy`, use types that know their size and make this correct by construction.

lib/mqtt/mqtt.cc
```diff
@@ -235,6 +226,58 @@ namespace
...
+	 */
+	constexpr const char clientIDCharacters[] =
+	  "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
+	constexpr const int maxClientIDLength = 23;
+
+	/**
+	 * Helper to check if a client ID is valid according to the MQTT
+	 * specification.
+	 */
+	bool is_valid_client_id(const char *id, size_t length)
```
davidchisnall, line 246: This is internal and so should take a `std::string_view`.

Then the code becomes:

```c++
if (id.size() > MQTTMaximumClientIDSize)
{
	return false;
}
for (c in id)
{
	switch (c)
	{
		default:
			return false;
		case '0' ... '9':
		case 'a' ... 'z':
		case 'A' ... 'Z':
			continue;
	}
}
```

lib/mqtt/mqtt.cc
```diff
@@ -235,6 +226,58 @@ namespace
...
+		// available
+		for (size_t i = 0; i < length; i++)
+		{
+			if (buffer[i] == '\0')
+			{
+				return true;
+			}
+		}
+		return false;
+	}
```
davidchisnall, line 279:
```suggestion
bool has_null_character(std::string_view buffer)
{
	return buffer.find('\0') != std::string_view::npos;
}
```

C++23 makes this simpler, but our stdlib is a bit too old.

lib/mqtt/mqtt.cc
```diff
@@ -471,6 +515,12 @@ SObj mqtt_connect(Timeout                    *t,
 		return nullptr;
 	}
 
+	if (!is_valid_client_id(clientID, clientIDLength))
+	{
+		Debug::log("Passed invalid client ID.");
+		return nullptr;
+	}
```
davidchisnall, line 522:
```suggestion
	if constexpr (DebugMQTT)
	{
		if (!is_valid_client_id(clientID, clientIDLength))
		{
			Debug::log("Passed invalid client ID.");
			return nullptr;
		}
	}
```

lib/mqtt/mqtt.cc
```diff
@@ -583,13 +633,12 @@ SObj mqtt_connect(Timeout                    *t,
 	// Prepare the low-level MQTT API connect call.
 	MQTTConnectInfo_t connectInfo = {0};
 
-	// We want to create a new session with the broker.
-	connectInfo.cleanSession = true;
-
+	connectInfo.cleanSession           = newSession;
 	connectInfo.pClientIdentifier      = clientID;
```
davidchisnall, line 637: Why not capture this as a std::string_view?  Does it need to remain valid across calls?  If so, we should document that.  If not, then we need to document that it does need to have global because, even though we're not capturing it, we are stashing it in a heap object.  Would it make sense to just copy this, since it's only 23 bytes?

lib/mqtt/mqtt.cc
```diff
@@ -758,6 +816,27 @@ int mqtt_publish(Timeout    *t,
 		return -EINVAL;
 	}
 
+	/**
```
davidchisnall, line 819: This probably also wants to be in an `if constexpr (DebugMQTT)` block.

lib/mqtt/mqtt.cc
```diff
@@ -806,6 +885,16 @@ int mqtt_subscribe(Timeout    *t,
 		return -EINVAL;
 	}
 
+	if (filterLength < 1)
```
davidchisnall, line 888: As does this.

lib/mqtt/mqtt.cc
```diff
@@ -854,6 +943,16 @@ int mqtt_unsubscribe(Timeout    *t,
...
```
davidchisnall, line 946: And this.

lib/mqtt/mqtt.cc
```diff
@@ -954,3 +1053,30 @@ int mqtt_run(Timeout *t, SObj mqttHandle)
...
+
+	if (!CHERI::check_pointer(buffer, length))
+	{
+		return -EINVAL;
+	}
+
+	for (size_t i = 0; i < length; i++)
+	{
+		// Select a character at random.
+		buffer[i] = characters[rand() % characters.size()];
```
davidchisnall, line 1075: Note that, since this is not a power of two, this will give a slightly biased distribution.  It's fine for avoiding collisions, but the API should document that fact.

lib/mqtt/mqtt.cc
```diff
@@ -954,3 +1053,30 @@ int mqtt_run(Timeout *t, SObj mqttHandle)
 		  return 0;
 	  });
 }
+
+int mqtt_generate_client_id(char *buffer, size_t length)
+{
+	auto characters =
+	  std::string_view(clientIDCharacters, sizeof(clientIDCharacters) - 1);
+
+	if (length > maxClientIDLength || !length)
```
davidchisnall, line 1062: Please don't use `!length` to check that an integer is equal to 0.  Also, please put brackets around any sequence of different operators.

```suggestion
	if ((length > maxClientIDLength) || (length == 0))
```

lib/tls/tls.cc
```diff
@@ -248,7 +248,11 @@ namespace
 		  t, connection->socket, inputBuffer, length);
 		Debug::log("Network stack returned {}", received);
 		// Any failure here is treated the same way: give up.
-		if (received < 0)
+		// Note: the BearSSL documentation says 'The len value MUST NOT
+		// be 0', so we treat it as an error too. Still, assuming
+		// `network_socket_receive_preallocated` is correctly
+		// implemented, 0 should never be returned.
+		if (received < 0 || received == 0 /* shoud not happen */)
```
davidchisnall, line 255: Why not <=?

examples/04.MQTT/mqtt.cc
```diff
@@ -114,6 +116,16 @@ void __cheri_compartment("mqtt_example") example()
 	auto heapAtStart =
 	  heap_quota_remaining(STATIC_SEALED_VALUE(mqttTestMalloc));
 
+	Debug::log("Generating client ID...");
+	// Prefix with something recognizable, for convenience. Note: do not
+	// copy \0, as this will be added by `mqtt_generate_client_id`.
+	memcpy(clientID, clientIDPrefix, clientIDPrefixLengthWithoutZero);
```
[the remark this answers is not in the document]
hlef: `-` is not a valid client ID character (though the validation would catch this :wink: )

But sure I can do something like that.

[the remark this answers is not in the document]
hlef: Actually - ignoring this given your other comments below.

examples/04.MQTT/mqtt.cc
hlef: Any suggestion to avoid the `memcpy`?

lib/mqtt/mqtt.cc
hlef: That's what I initially had, but it doesn't link:
```
error: ld.lld: error: undefined symbol: memchr
>>> referenced by debug.hh:588 (/mnt/workspace/cheriot-rtos/sdk/include/debug.hh:588)
>>>               build/cheriot/cheriot/release/MQTT.compartment:()
```
We don't have `memchr`, hence the comment.

lib/mqtt/mqtt.cc
hlef: The client ID is only needed for the connect. It does not need to be valid across calls. `connectInfo` is not a heap object, it's on the stack?

I suggest to leave things as they are and just document the lifetime of the client ID in the API.

hlef:
@davidchisnall Thanks for the comments! I addressed all I could, answered to those where more discussion is needed.

lib/mqtt/mqtt.cc
davidchisnall: We should probably just add it, it’s very small.

hlef: Can do. But how about we merge this PR first, I add `memchr` ASAP, and go back to this code once it's in? I don't want to delay the merge of this PR by a week if possible

davidchisnall (approved):

PR #10, opened 2024-03-25
Fix `FreeRTOS_connect` return value handling.

hlef:
`-pdFREERTOS_ERRNO_EISCONN` actually means success, so we should consider it like 0 when processing the return value (instead of returning `-EINVAL`.

This fix was initially proposed by David in an experimental branch.

While at it, put the `default` block at the end because it is the least likely to execute (vast majority of cases will be success or timeout).

### DIFF
--- a/lib/tcpip/network_wrapper.cc
+++ b/lib/tcpip/network_wrapper.cc
@@ -496,7 +496,8 @@ int network_socket_connect_tcp_internal(Timeout       *timeout,
 		  {
 			  default:
 				  return -EINVAL;
-			  case 0:
+			  case 0: // success
+			  case -pdFREERTOS_ERRNO_EISCONN: // already connected
 				  Debug::log("Successfully connected to server");
 				  return 0;
 			  case -pdFREERTOS_ERRNO_EWOULDBLOCK:


davidchisnall:
> While at it, put the default block at the end because it is the least likely to execute (vast majority of cases will be success or timeout).

Please don't do this.  Putting `default` first is a common defensive programming idiom to prevent future changes from accidentally falling through into the `default` case.  Any `switch` that has a `default` that is not first needs extra review.

The order of things in a switch does not imply probability.  It will not generate branch-weight probabilities.  There are annotations that we can add if we need that.

hlef:
> > While at it, put the default block at the end because it is the least likely to execute (vast majority of cases will be success or timeout).
> 
> Please don't do this. Putting `default` first is a common defensive programming idiom to prevent future changes from accidentally falling through into the `default` case. Any `switch` that has a `default` that is not first needs extra review.
> 
> The order of things in a switch does not imply probability. It will not generate branch-weight probabilities. There are annotations that we can add if we need that.

Interesting. I didn't mean it in terms of branch-weight probabilities, but rather in terms of readability. For me, the first cases should be the most common ones, followed by error cases, concluding with cases that basically should never happen.

davidchisnall (approved):

PR #12, opened 2024-03-28
Various fixes to TCP/IP and TLS

hlef:
This PR contains various fixes to TCP/IP and TLS, partly taken and cleaned up from the experimental MQTT demo branch.

Individual commits provide detailed descriptions of the changes.

### DIFF
--- a/lib/tcpip/BufferManagement.cc
+++ b/lib/tcpip/BufferManagement.cc
@@ -10,6 +10,26 @@
 
 using Debug = ConditionalDebug<false, "Buffer management">;
 
+// Use a separate allocator quota for the buffer manager (false by default).
+// The buffer manager is responsible for allocating network buffers, which
+// differs from the other types of allocations the TCP/IP stack performs. It
+// may thus make sense to give it its own quota. We may want to expose this as
+// a build system configuration option at some point.
+#define USE_DEDICATED_BUFFERMANAGER_POOL false
+// Size of the buffer manager allocator quota. Only relevant if
+// USE_DEDICATED_BUFFERMANAGER_POOL is enabled.
+#define BUFFER_MANAGER_QUOTA                                                   \
+	(ipconfigNUM_NETWORK_BUFFER_DESCRIPTORS *                                  \
+	 (ipconfigTCP_MSS + ipBUFFER_PADDING))
+
+#if USE_DEDICATED_BUFFERMANAGER_POOL
+DECLARE_AND_DEFINE_ALLOCATOR_CAPABILITY(BufferManagementMallocQuota,
+                                        BUFFER_MANAGER_QUOTA);
+#	define BM_MALLOC_QUOTA STATIC_SEALED_VALUE(BufferManagementMallocQuota)
+#else
+#	define BM_MALLOC_QUOTA MALLOC_CAPABILITY
+#endif
+
 constexpr size_t MinimumBufferSize =
 #if ipconfigUSE_TCP == 1
   sizeof(TCPPacket_t)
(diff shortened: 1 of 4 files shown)


lib/tcpip/BufferManagement.cc
```diff
@@ -10,6 +10,23 @@
...
+// may thus make sense to give it its own quota. We may want to expose this as
+// a build system configuration option at some point.
+#define USE_DEDICATED_BUFFERMANAGER_POOL false
+// Size of the buffer manager allocator quota. Only relevant if
+// USE_DEDICATED_BUFFERMANAGER_POOL is enabled.
+#define BUFFER_MANAGER_QUOTA (ipconfigNUM_NETWORK_BUFFER_DESCRIPTORS * (ipconfigTCP_MSS + ipBUFFER_PADDING))
+
+#if USE_DEDICATED_BUFFERMANAGER_POOL
+DECLARE_AND_DEFINE_ALLOCATOR_CAPABILITY(BMMalloc, BUFFER_MANAGER_QUOTA);
+#define BM_MALLOC STATIC_SEALED_VALUE(BMMalloc)
```
davidchisnall, line 25: Please name this as quota or capability. I would assume that something called foo malloc is a malloc for foo, not an object passed to an allocator.

lib/tcpip/BufferManagement.cc
```diff
@@ -49,16 +66,19 @@ pxGetNetworkBufferWithDescriptor(size_t     xRequestedSizeBytes,
 
 	Timeout t{xBlockTimeTicks};
 
-	std::unique_ptr<NetworkBufferDescriptor_t> descriptor{
+	// TODO we likely want to pre-allocate (or re-use) the descriptors at
+	// some point
+	auto deleter = [=](void *ptr) { heap_free(BM_MALLOC, ptr); };
```
davidchisnall, line 71: Can you check that this is hitting the special cases in unique ptr for deletes that are pure functions (and so don’t need to be allocated)?

lib/tls/tls.cc
```diff
@@ -318,10 +319,10 @@ namespace
 				  {
 					  return -ENOTCONN;
 				  }
-				  // If there are data ready to receive, return
-				  // it immediately.
-				  if ((state & BR_SSL_RECVAPP) == BR_SSL_RECVAPP)
+				  else if ((state & BR_SSL_RECVAPP) == BR_SSL_RECVAPP)
```
davidchisnall, line 322: Clang-tidy should complain about this, please make sure that you run it.

lib/tls/tls.cc
```diff
@@ -349,16 +350,18 @@ namespace
 					    "Received {} bytes into {}", length, receivedBuffer);
 					  return ssize_t(length);
 				  }
-				  if ((state & BR_SSL_RECVREC) == BR_SSL_RECVREC)
+				  else if ((state & BR_SSL_RECVREC) == BR_SSL_RECVREC)
```
davidchisnall, line 353: And this. No else after return.

lib/tls/tls.cc
```diff
@@ -349,16 +350,18 @@ namespace
...
-						  Debug::log("Connection closed, shutting down");
-						  return 0;
+						  // The link died. After
+						  // getting -ENOTCONN, the
+						  // caller should close the
+						  // TLS socket.
+						  return -ENOTCONN;
 					  }
-					  if (received == -ETIMEDOUT)
+					  else if (received == -ETIMEDOUT)
```
davidchisnall, line 364: As above.

lib/tls/tls.cc
```diff
@@ -587,8 +598,9 @@ ssize_t tls_connection_send(Timeout *t,
 					  thread_sleep(&shortSleep);
 					  t->elapse(shortSleep.elapsed);
 				  }
-				  if (!t->may_block())
+				  else
```
davidchisnall, line 601: Are you sure about this? The line above may have reached the timeout condition.

lib/tls/tls.cc
davidchisnall: I have now added it to CI.

lib/tls/tls.cc
hlef: Right. The two ifs in a row are correct but this needs a comment.

lib/tcpip/BufferManagement.cc
hlef: Interesting. I am investigating that.

hlef: I have looked at different approaches to deleters (default deleter, lambda, std::function, deleter class), and curiously they all seem to have the same memory footprint (8 Bytes on top of the allocation itself).

This seems to contradict some of the stuff I've found online (https://alexanderhoughton.co.uk/blog/c-unique-shared-ptr-custom-deleters/), but we have a different compiler and it's been a few years since the article was published.

PR #15, opened 2024-04-03
Three bug fixes in the TCP/IP stack.

hlef:
This PR fixes three issues in the TCP/IP stack:

- SNTP is broken because of a missing allocator capability
- The custom malloc quota (off by default) of the buffer manager is also broken
- DHCP lease renew does not work because the firewall filters out the server's packets


lib/firewall/firewall.cc
```diff
@@ -12,6 +12,19 @@
 #include <timeout.hh>
 #include <vector>
 
+namespace
+{
+	// TODO These two should probably be in their own library.
+	uint16_t ntohs(uint16_t value)
+	{
+		return __builtin_bswap16(value);
```
davidchisnall, line 20: We currently support only little endian (and probably will never support big endian), but these should probably be wrapped in a check for `__LITTLE_ENDIAN__`.  They should probably also be `constexpr`.

hlef:
Alright, I have tested the fix for the DHCP issue and everything is good now.

For the record, this patch to `third_party/freertos-plus-tcp` is sufficient to reliably trigger the DHCP lease renewal issue with a high rate in the demo:
```
diff --git a/source/FreeRTOS_DHCP.c b/source/FreeRTOS_DHCP.c
index 5ebeebc..632fd4d 100644
--- a/source/FreeRTOS_DHCP.c
+++ b/source/FreeRTOS_DHCP.c
@@ -1092,7 +1092,8 @@
 
                     /* The DHCP parameter is in seconds, convert
                      * to host-endian format. */
-                    EP_DHCPData.ulLeaseTime = FreeRTOS_ntohl( pxSet->ulParameter );
+                    //EP_DHCPData.ulLeaseTime = FreeRTOS_ntohl( pxSet->ulParameter );
+                    EP_DHCPData.ulLeaseTime = 10;
 
                     /* Divide the lease time by two to ensure a renew
                      * request is sent before the lease actually expires. */
diff --git a/source/include/FreeRTOS_DHCP.h b/source/include/FreeRTOS_DHCP.h
index 26fef7d..e0e6c7f 100644
--- a/source/include/FreeRTOS_DHCP.h
+++ b/source/include/FreeRTOS_DHCP.h
@@ -108,12 +108,10 @@
 //#define EP_DHCPData                     xDHCPData              /**< Temporary define to make /single source similar to /multi version. */
 //#define EP_IPv4_SETTINGS                xNetworkAddressing     /**< Temporary define to make /single source similar to /multi version. */
 
-/** @brief If a lease time is not received, use the default of two days (48 hours in ticks).
- * Can not use pdMS_TO_TICKS() as integer overflow can occur. */
-#define dhcpDEFAULT_LEASE_TIME          ( ( 48UL * 60UL * 60UL ) * configTICK_RATE_HZ )
+#define dhcpDEFAULT_LEASE_TIME          ( pdMS_TO_TICKS( 10000UL ) )		/* 10 seconds in ticks. */
 
 /** @brief Don't allow the lease time to be too short. */
-#define dhcpMINIMUM_LEASE_TIME          ( pdMS_TO_TICKS( 60000UL ) )            /* 60 seconds in ticks. */
+#define dhcpMINIMUM_LEASE_TIME          ( pdMS_TO_TICKS( 10000UL ) )            /* 10 seconds in ticks. */
 
 /** @brief Marks the end of the variable length options field in the DHCP packet. */
 #define dhcpOPTION_END_BYTE             0xffu
```

lib/firewall/firewall.cc
```diff
@@ -12,6 +12,25 @@
 #include <timeout.hh>
 #include <vector>
 
+namespace
+{
+#ifdef __LITTLE_ENDIAN__
```
davidchisnall, line 17: I meant in the bodies of these. They should still exist on big endian, they should just return their arguments.

davidchisnall (approved):

PR #18, opened 2024-04-09
Reduce code size of the firewall.

hlef:
The firewall is a little bit too big. Reduce the code size by...

- Removing the template in `EndpointsTable`. Instead, go for a fixed-size buffer that can contain both IPV4 and IPV6 addresses.
- Simplify `find_endpoint` and `remove_endpoint` which rely on large templated C++ utility functions.

With this, we gain about 2kB.

### DIFF
--- a/lib/firewall/firewall.cc
+++ b/lib/firewall/firewall.cc
@@ -175,8 +175,56 @@ namespace
 
 	static_assert(sizeof(IPv4Header) == 20);
 
+	static constexpr const uint8_t Ipv6AddressSize = 16;
+
 	/**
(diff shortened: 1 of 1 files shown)


davidchisnall:
This wastes a lot of space for IPv4 addresses. I would still like separate arrays for each, I just want them sorted / searched with something like qsort (sort variable-sized things taking the size as an argument, using memcmp rather than comparison operators for sorting).  This can be done either by writing in a more C style, or by providing iterations to the standard algorithms that are initialised with a dynamic size.

hlef:
Sounds like we want a different approach here. I'll close this one for now and come back with what we discussed.

PR #22, opened 2024-04-19
Rework `network_socket_close` error paths.

hlef:
49f42c1e1f7fcbacace0bcd52a58f98013168616 forgot to add unlocks on error paths in `network_socket_close`. Address this.

At the same time, clarify that we have two types of errors here: those where the call can be retried (e.g., called with the wrong malloc capability, or an invalid timeout), and those which cannot be retried (the socket close failed due to an unspecified error and we went ahead to free it). Both are currently under `-EAGAIN`. I think that we should differentiate the two to give a chance to the caller to call again with the right arguments. Address this by moving unrecoverable errors to `-ENOTRECOVERABLE` and update the documentation accordingly.

Finally, a socket close return value of 0 is unrecoverable and most likely due to an internal error, so in that case we should go ahead and free everything instead of returning `-EAGAIN`.

**Note**: Since all errors currently under `-ENOTRECOVERABLE` are internal, not supposed to happen (modulo bug), and not acted upon (we're leaking memory, but there really isn't anything we can do about that apart from maybe reseting, once we have it working?), we could also simply move them as debug asserts and ignore the failure.

### DIFF
--- a/lib/tcpip/network_wrapper.cc
+++ b/lib/tcpip/network_wrapper.cc
@@ -558,87 +558,95 @@ int network_socket_close(Timeout *t, SObj mallocCapability, SObj sealedSocket)
 	}
 	return with_sealed_socket(
 	  [=](SealedSocket *socket) {
-		  // Don't use a lock guard here, we don't want to release the lock if
-		  // it's been deallocated.  This must be released on all return paths
-		  // except the one for success.
-		  if (!socket->socketLock.try_lock(t))
+		  if (LockGuard g{socket->socketLock, t})
 		  {
-			  return -ETIMEDOUT;
-		  }
-		  // Since we free the socket and the token at the end after
-		  // terminating the socket, ensure that the frees won't fail
-		  if (heap_can_free(mallocCapability, socket->socket) != 0 ||
-		      token_obj_can_destroy(
-		        mallocCapability, socket_key(), sealedSocket) != 0)
-		  {
-			  return -EINVAL;
-		  }
-		  bool isTCP = socket->socket->ucProtocol == FREERTOS_IPPROTO_TCP;
-		  // Shut down the socket before closing the firewall.  Don't
-		  // bother with the return value: `FreeRTOS_shutdown` fails
-		  // mainly if the TCP connection is dead, which is likely to
-		  // happen in practice and has no impact for us.
-		  FreeRTOS_shutdown(socket->socket, FREERTOS_SHUT_RDWR);
-		  auto localPort = ntohs(socket->socket->usLocalPort);
-		  if (socket->socket->bits.bIsIPv6)
-		  {
-			  if (isTCP)
+			  // Since we free the socket and the token at the end after
+			  // terminating the socket, ensure that the frees won't fail
+			  if (heap_can_free(mallocCapability, socket->socket) != 0 ||
+			      token_obj_can_destroy(
+			        mallocCapability, socket_key(), sealedSocket) != 0)
 			  {
-				  firewall_remove_tcpipv6_endpoint(localPort);
+				  Debug::log("Unable to free socket or token.");
+				  // The main reason why this would fail is because we
+				  // were called with the wrong malloc capability. We
+				  // want to leave a chance to the caller to call us
+				  // again with the right capability.
+				  return -EINVAL;
+			  }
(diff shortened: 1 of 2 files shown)


lib/tcpip/network_wrapper.cc
```diff
@@ -604,30 +611,34 @@ int network_socket_close(Timeout *t, SObj mallocCapability, SObj sealedSocket)
...
 		  auto closeStatus = close_socket_retry(t, socket->socket);
 		  if (closeStatus == 0)
 		  {
-			  return -EINVAL;
+			  // The only reason why this would fail is internal
+			  // corruption (did someone already close the
+			  // socket?). Nothing can be done by anyone at this
+			  // stage. Don't return because we would leak
+			  // everything.
+			  ret = -ENOTRECOVERABLE;
```
davidchisnall, line 623: I think here we should still free the memory, but report that something went wrong?

lib/tcpip/network_wrapper.cc
```diff
@@ -604,30 +611,34 @@ int network_socket_close(Timeout *t, SObj mallocCapability, SObj sealedSocket)
...
 		  {
 			  // This is not supposed to happen, since we did a
 			  // heap_can_free earlier. If it does, we're leaking
 			  // the socket.
 			  Debug::log("Failed to free socket.");
-			  // Release the lock so that we don't leak it.
-			  socket->socketLock.unlock();
 			  // Don't return now, try to at least free the token.
-			  ret = -EINVAL;
+			  ret = -ENOTRECOVERABLE;
```
davidchisnall, line 641: I believe the only way to reach this is to have insufficient stack to call `heap_free`.  Is there any other way that it can happen?

lib/tcpip/network_wrapper.cc
```diff
@@ -636,7 +647,7 @@ int network_socket_close(Timeout *t, SObj mallocCapability, SObj sealedSocket)
 			  // token_obj_can_destroy earlier. If it does, we're
 			  // leaking the token.
 			  Debug::log("Failed to free token.");
-			  ret = -EINVAL;
+			  ret = -ENOTRECOVERABLE;
```
davidchisnall, line 650: Same question as above.

lib/tcpip/network_wrapper.cc
hlef: Agree. Of course there is also the possibility that the caller frees it concurrently as it owns the heap capability but this has no impact here. And there is also the possibility of a bug somewhere, but in theory it shouldn't happen.

lib/tcpip/network_wrapper.cc
```diff
@@ -558,9 +558,9 @@ int network_socket_close(Timeout *t, SObj mallocCapability, SObj sealedSocket)
 	}
 	return with_sealed_socket(
 	  [=](SealedSocket *socket) {
-		  // Don't use a lock guard here, we don't want to release the lock if
-		  // it's been deallocated.  This must be released on all return paths
-		  // except the one for success.
+		  // Don't use a lock guard here, we don't want to release the
```
davidchisnall, line 561: We should probably add a `reset` method to the lock guard to swap the lock that is wraps (optionally with no lock) so that we can default to unlocking on all return paths but release lock ownership on the one where we've got rid of everything.

lib/tcpip/network_wrapper.cc
davidchisnall: I'm still a bit confused.  Doesn't this leak the socket object?  We've not freed everything, but we've told the caller not to retry?

lib/tcpip/network_wrapper.cc
davidchisnall: We can check for the former case by checking `-ECOMPARTMENTFAIL` on the `heap_free` return and the latter by checking the tag bit.  

In the case where there's a concurrent free, we should continue freeing the other things.

hlef: > In the case where there's a concurrent free, we should continue freeing the other things.

This is already what we do?

> We can check for the former case by checking -ECOMPARTMENTFAIL on the heap_free return and the latter by checking the tag bit.

What should we check it for exactly? To provide a more explicit error code?

hlef:
I have fixed the comments. Note that this will fail until we merge https://github.com/microsoft/cheriot-rtos/pull/213

davidchisnall (approved):

PR #23, opened 2024-05-16
Better handle TCP/IP stack crashes in the TLS compartment.

hlef:
When the TCP/IP stack crashes, API calls to the compartment return `-ECOMPARTMENTFAIL`. These should be treated similarly to `-ENOTCONN`.

Currently `-ECOMPARTMENTFAIL` failures are not considered by the TLS compartment and are handled in various (incorrect) ways across the code-base. Address this.

While we are at it, avoid the call to `network_socket_close` if the socket is a `nullptr`.

### DIFF
--- a/include/NetAPI.h
+++ b/include/NetAPI.h
@@ -163,6 +163,7 @@ NetworkReceiveResult __cheri_compartment("TCPIP")
  *
  * The negative values will be errno values:
  *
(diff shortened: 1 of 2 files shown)


hlef:
@davidchisnall Updated to address the comments.

davidchisnall (approved):

PR #24, opened 2024-05-16
Better handle TCP/IP stack crashes in the NETAPI compartment.

hlef:
When the TCP/IP stack crashes, API calls to the compartment return `-ECOMPARTMENTFAIL`.

Currently `-ECOMPARTMENTFAIL` failures are not considered by the NetAPI compartment. This leads to erratic behavior when the network stack crashes, such as adding nonsensical entries to the firewall.

Address this with proper checking.

### DIFF
--- a/lib/netapi/NetAPI.cc
+++ b/lib/netapi/NetAPI.cc
@@ -55,18 +55,37 @@ SObj network_socket_connect_tcp(Timeout *timeout,
 		Debug::log("Host capability does not authorise a TCP connection");
 		return nullptr;
 	}
+
+	NetworkAddress    address{NetworkAddress::AddressKindInvalid};
+	CHERI::Capability addressPtr = &address;
+	addressPtr.permissions() &= {CHERI::Permission::Store};
 	firewall_permit_dns();
-	NetworkAddress address = network_host_resolve(host->hostname, UseIPv6);
+	int ret = network_host_resolve(host->hostname, UseIPv6, addressPtr);
 	firewall_permit_dns(false);
-	if (address.kind == NetworkAddress::AddressKindInvalid)
+	if ((ret < 0) || (address.kind == NetworkAddress::AddressKindInvalid))
 	{
 		Debug::log("Failed to resolve host");
 		return nullptr;
 	}
-	bool isIPv6       = address.kind == NetworkAddress::AddressKindIPv6;
-	auto sealedSocket = network_socket_create_and_bind(
+	bool isIPv6 = address.kind == NetworkAddress::AddressKindIPv6;
+
+	CHERI::Capability sealedSocket = network_socket_create_and_bind(
 	  timeout, mallocCapability, isIPv6, ConnectionTypeTCP);
-	auto kind = network_socket_kind(sealedSocket);
+	if (!sealedSocket.is_valid())
+	{
+		Debug::log("Failed to create socket");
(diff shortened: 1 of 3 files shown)


lib/netapi/NetAPI.cc
```diff
@@ -55,18 +55,43 @@ SObj network_socket_connect_tcp(Timeout *timeout,
...
 	NetworkAddress address = network_host_resolve(host->hostname, UseIPv6);
 	firewall_permit_dns(false);
-	if (address.kind == NetworkAddress::AddressKindInvalid)
+	if (address.kind == NetworkAddress::AddressKindIPv6)
 	{
+		isIPv6 = true;
+	}
+	else if (address.kind != NetworkAddress::AddressKindIPv4)
+	{
+		// Do not check for `NetworkAddress::AddressKindInvalid`
```
davidchisnall, line 69: This is currently fragile. We need a compiler issue to track the fact that we’re dead-store eliminating initialisation of sret parameters for cross-compartment calls. Then the original code would be correct. The neq code is correct only if we happen to have specific stack contents at the start.

lib/netapi/NetAPI.cc
```diff
@@ -55,18 +55,43 @@ SObj network_socket_connect_tcp(Timeout *timeout,
...
+		// during the call to `network_host_resolve`, returning
+		// `-ECOMPARTMENTFAIL`.
 		Debug::log("Failed to resolve host");
 		return nullptr;
 	}
-	bool isIPv6       = address.kind == NetworkAddress::AddressKindIPv6;
+
 	auto sealedSocket = network_socket_create_and_bind(
 	  timeout, mallocCapability, isIPv6, ConnectionTypeTCP);
+	if (__builtin_cheri_tag_get(sealedSocket) == false)
```
davidchisnall, line 79: We’re in C++ here, so just make sealedSocket a Capability.

lib/netapi/NetAPI.cc
```diff
@@ -55,18 +55,43 @@ SObj network_socket_connect_tcp(Timeout *timeout,
...
+	{
+		Debug::log("Failed to create socket");
+		return nullptr;
+	}
+
 	auto kind = network_socket_kind(sealedSocket);
+	// Here again, do not check for `Invalid` because this will miss cases
```
davidchisnall, line 86: As above, this is an incorrect work around for a compiler bug.

davidchisnall (commented):
Most of this is incorrectly working around compiler bugs, we should fix the bugs first.

hlef:
@davidchisnall Updated the pull request to address these issues.

lib/tcpip/network-internal.h
```diff
@@ -17,9 +17,14 @@
...
+ *
+ * The result of the resolve is stored in `address`.
+ *
+ * This returns zero for success, or a negative value on error.
  */
-NetworkAddress __cheri_compartment("TCPIP")
-  network_host_resolve(const char *hostname, bool useIPv6);
+__cheri_compartment("TCPIP") int network_host_resolve(const char     *hostname,
+                                                      bool            useIPv6,
+                                                      NetworkAddress *address);
```
davidchisnall, line 27:
```suggestion
                                                      NetworkAddress *outAddress);
```

davidchisnall (approved):

PR #26, opened 2024-05-22
All of the fixes needed for the demo to work

davidchisnall:


lib/mqtt/mqtt.cc
```diff
@@ -68,7 +68,7 @@ namespace
 		NetworkContext_t     networkContext;
 
 		// Lock on which the whole public API synchronizes.
-		FlagLockPriorityInherited lock;
+		RecursiveMutex lock;
```
hlef, line 71: I think that we should add a comment here to explain that CoreMQTT officially discourages recursively calling into the API, and that users recursively calling into the MQTT compartment should be doing it at their own risk (t works on the normal path but we have not asserted that all error paths are safe).

Maybe linking to this issue? https://github.com/FreeRTOS/coreMQTT/issues/278

lib/mqtt/mqtt.cc
```diff
@@ -534,7 +541,7 @@ SObj mqtt_connect(Timeout                    *t,
 	// coreMQTT), we can assume that the allocator zeroes out for us.
 	size_t handleSize =
 	  sizeof(CHERIoTMqttContext) -
-	  sizeof(CHERIoTMqttContext::variableLengthData) + networkBufferSize +
+	  sizeof(CHERIoTMqttContext::variableLengthData) +
```
hlef, line 544: If doing this we need to update the comment "Following this we allocate variable length data" at the end of the `CHERIoTMqttContext` object (it still mentions "the network buffer (array of uint8_t)")

include/mqtt.h
```diff
@@ -257,6 +257,12 @@ int __cheri_compartment("MQTT") mqtt_unsubscribe(Timeout    *t,
  * Fetch ACK and PUBLISH notifications on a given MQTT connection, and keep
  * the connection alive.
  *
+ * This function will invoke the callbacks passed to `mqtt_connect`.  The
+ * connection object is protected by a recursive mutex, so these callbacks can
+ * call additional publish and subscribe functions.  If doing so, care must be
+ * taken to ensure that the buffer is not exhausted.  Calling `mqtt_run` from a
```
hlef, line 263: Not sure what "care must be taken to ensure that the buffer is not exhausted" means here? This buffer is neither passed by users nor exposed to users, so it would be nice to say how users can avoid exhausting it. Does this mean limiting the amount of recursive calls?

lib/mqtt/mqtt.cc
```diff
@@ -575,8 +582,14 @@ SObj mqtt_connect(Timeout                    *t,
 	  reinterpret_cast<MQTTPubAckInfo_t *>(&context->variableLengthData);
 	MQTTPubAckInfo_t *outgoingPublishes =
 	  incomingPublishes + incomingPublishCount;
-	uint8_t *networkBuffer = reinterpret_cast<uint8_t *>(outgoingPublishes) +
-	                         sizeof(MQTTPubAckInfo_t) * outgoingPublishCount;
+	uint8_t *networkBuffer =
+	  static_cast<uint8_t *>(heap_allocate(t, allocator, networkBufferSize));
+
+	if (networkBuffer == nullptr)
```
hlef, line 588: Since we are currently fixing this... it would be nice to check the tag bit instead of `nullptr` here?

lib/mqtt/mqtt.cc
```diff
@@ -575,8 +582,14 @@ SObj mqtt_connect(Timeout                    *t,
 	  reinterpret_cast<MQTTPubAckInfo_t *>(&context->variableLengthData);
 	MQTTPubAckInfo_t *outgoingPublishes =
 	  incomingPublishes + incomingPublishCount;
-	uint8_t *networkBuffer = reinterpret_cast<uint8_t *>(outgoingPublishes) +
-	                         sizeof(MQTTPubAckInfo_t) * outgoingPublishCount;
+	uint8_t *networkBuffer =
```
hlef, line 585: Nit but for future readers it may be nice to add a comment here to tell that we need a separate allocation because the total size of a sealed object is limited and embedding this into the `CHERIoTMqttContext` object we allocated earlier would prevent us from sending/receiving large payloads.

lib/firewall/firewall.cc
```diff
@@ -479,19 +521,24 @@ namespace
...
 			case EtherType::IPv6:
 			{
 				Debug::log("Permitting outbound IPv6 packet");
 				return true;
 				break;
 			}
+#endif
 		}
-		return true;
+		return false;
```
hlef, line 535: This should never be reached right?

hlef (changes requested):
Looks good, just a few nits.

lib/mqtt/mqtt.cc
nwf-msr: This could be reworked now that https://github.com/microsoft/cheriot-rtos/pull/244 has landed. 🎉

lib/firewall/firewall.cc
```diff
@@ -786,9 +853,17 @@ bool ethernet_driver_start(std::atomic<uint8_t> *state)
...
 	barrier = 2;
 	barrier.notify_one();
 	return true;
 }
+
+uint8_t *firewall_mac_address_get()
+{
+	CHERI::Capability ret{mac_address().data()};
+	ret.permissions() &= {CHERI::Permission::Load, CHERI::Permission::Global};
+	ret.bounds() = 6;
```
nwf-msr, line 867: Just for my own edification, what are the bounds absent this line?

davidchisnall: Looks like it's 6.  I'll change this to an assertion.

lib/mqtt/mqtt.cc
davidchisnall: This is gone now that the RTOS doesn't impose this restriction.

lib/firewall/firewall.cc
```diff
@@ -532,6 +572,19 @@ namespace
 		}
 		EthernetHeader *ethernetHeader =
 		  reinterpret_cast<EthernetHeader *>(const_cast<uint8_t *>(data));
+		if ((ethernetHeader->destination != mac_address()) &&
+		    (ethernetHeader->destination != broadcastMAC))
```
nwf-msr, line 576: Perhaps a comment that we don't yet support multicast?

PR #27, opened 2024-06-14
Restart the TCP/IP stack on crash.

hlef:
The TCP/IP stack currently has a lot of state and cannot recover gracefully in case of failure. This commit addresses this limitation.

We introduce a custom error handler for the TCP/IP stack. At a high-level, this error handler terminates all user threads currently present in the TCP/IP compartment and resets the IP thread to its entry point `ip_thread_entry`. It does so by setting all the locks of the compartment for destruction and notifying all futex waiters. Then it frees all heap allocations and resets globals. After all the state has been cleaned up, it restarts the network stack.

**At this stage, this is more of a RFC than a proper PR for merge.** Some documentation TODOs are still there, which I will process in the coming days.

This PR was tested as following:

- *Network thread crash*: A crash of the network thread can conveniently be triggered through a ping with the following patch to FreeRTOS+TCP:

```
diff --git a/source/FreeRTOS_ICMP.c b/source/FreeRTOS_ICMP.c
index 78cc9c6..226866f 100644
--- a/source/FreeRTOS_ICMP.c
+++ b/source/FreeRTOS_ICMP.c
@@ -71,6 +71,8 @@
 
 #if ( ipconfigREPLY_TO_INCOMING_PINGS == 1 ) || ( ipconfigSUPPORT_OUTGOING_PINGS == 1 )
 
+static int counter = 0;
+
 /**
  * @brief Process an ICMP packet. Only echo requests and echo replies are recognised and handled.
  *
@@ -98,6 +100,16 @@
             /* coverity[misra_c_2012_rule_11_3_violation] */
             ICMPPacket_t * pxICMPPacket = ( ( ICMPPacket_t * ) pxNetworkBuffer->pucEthernetBuffer );
 
+       if (counter > 0)
+       {
+               printf("(!) triggering a crash\n");
+               *((volatile int *) 0x0) = 0;
+       }
+       else
+       {
+               counter++;
+       }
+
             switch( pxICMPPacket->xICMPHeader.ucTypeOfMessage )
             {
                 case ipICMP_ECHO_REQUEST:
```

   Note that this ignores the first ping received by the system as one ping may be sent by the network in the first boot phase. You may thus need to send two pings to trigger the crash depending on your network setup. Feel free to adjust this.

- *User thread crash*: A tricky user thread crash can be triggered in the critical section of the sealed socket locks with the following patch:

```
diff --git a/lib/tcpip/network_wrapper.cc b/lib/tcpip/network_wrapper.cc
index 00510a6..8201159 100644
--- a/lib/tcpip/network_wrapper.cc
+++ b/lib/tcpip/network_wrapper.cc
@@ -483,6 +483,7 @@ int network_host_resolve(const char     *hostname,
 	  -1 /* invalid if we are restarting */);
 }
 
+static int counter = 0;
 SObj network_socket_create_and_bind(Timeout       *timeout,
                                     SObj           mallocCapability,
                                     bool           isIPv6,
@@ -562,6 +563,16 @@ SObj network_socket_create_and_bind(Timeout       *timeout,
 			  localAddress.sin_family = Family;
 			  auto bindResult =
 			    FreeRTOS_bind(socket, &localAddress, sizeof(localAddress));
+			  if (counter == 1)
+			  {
+				  counter++;
+				  printf("(!) triggering a crash in socket create\n");
+				  *((volatile int *)0x0) = 0;
+			  }
+			  else
+			  {
+				  counter++;
+			  }
 			  if (bindResult != 0)
 			  {
 				  // See above comments.
```

   The TCP/IP stack should detect that the crash happened with the lock held, forcefully unlock it, and continue its business.

**Unmerged dependencies in the main RTOS tree.** We need the following two PRs, which will be merged soon:
- https://github.com/microsoft/cheriot-rtos/pull/249
- https://github.com/microsoft/cheriot-rtos/pull/248

**Known issue:**
- The timeout of `malloc` is still not quite right: https://github.com/microsoft/cheriot-rtos/blob/main/sdk/include/stdlib.h#L96

   The value of 30 is generally enough, however in one instance it was still too low and triggered a hang during reset. We should consider increasing this value at a later point or find a more bullet-proof solution. You can easily detect such hangs with the following patch to FreeRTOS+TCP:

```
diff --git a/source/FreeRTOS_TCP_WIN.c b/source/FreeRTOS_TCP_WIN.c
index 8296657..65a02cb 100644
--- a/source/FreeRTOS_TCP_WIN.c
+++ b/source/FreeRTOS_TCP_WIN.c
@@ -430,6 +430,7 @@
 
             if( xTCPSegments == NULL )
             {
+                printf("Allocation of sectors failed.\n");
                 FreeRTOS_debug_printf( ( "prvCreateSectors: malloc %u failed\n",
                                          ( unsigned ) ( ipconfigTCP_WIN_SEG_COUNT * sizeof( xTCPSegments[ 0 ] ) ) ) );
```

### DIFF
--- a/lib/firewall/firewall.cc
+++ b/lib/firewall/firewall.cc
@@ -1,7 +1,7 @@
 // Copyright SCI Semiconductor and CHERIoT Contributors.
 // SPDX-License-Identifier: MIT
 
-#include "firewall.h"
+#include "firewall.hh"
 #include <atomic>
 #include <compartment-macros.h>
 #include <debug.hh>
@@ -68,6 +68,11 @@ namespace
 	}
 
 	std::atomic<uint32_t> barrier;
+
+	/**
+	 * This is used to synchronize with the TCP/IP stack during resets.
+	 */
+	std::atomic<uint8_t> *tcpipRestartState = nullptr;
 
 	auto &lazy_network_interface()
 	{
@@ -238,6 +243,13 @@ namespace
 		{
 			static EndpointsTable table;
 			return table;
+		}
+
+		void clear(IPProtocolNumber protocol)
+		{
+			auto guardedTable = permitted_endpoints(protocol);
+			auto &[g, table]  = guardedTable;
+			table.clear();
 		}
 
 		void remove_endpoint(IPProtocolNumber protocol,
@@ -492,6 +504,16 @@ namespace
 
 	bool packet_filter_ingress(const uint8_t *data, size_t length)
 	{
+		uint32_t stateSnapshot = tcpipRestartState->load();
+		if (stateSnapshot != 0 &&
+		    ((stateSnapshot & RestartStateDriverKickedBit) == 0))
+		{
+			// We are in a reset and the driver has not yet been
+			// restarted.
+			Debug::log("Dropping packet due to network stack restart.");
+			return false;
+		}
+
 		// Not a valid Ethernet frame (64 bytes including four-byte FCS, which
 		// is stripped by this point).
 		if (length < 60)
@@ -526,26 +548,6 @@ namespace
 	std::atomic<uint32_t> receivedCounter;
 
 } // namespace
-
-bool ethernet_driver_start()
-{
-	// Protect against double entry.  If the barrier state is 0, no
-	// initialisation has happened and we should proceed.  If it's 1, we're in
-	// the middle of initialisation, if it's 2 then initialisation is done.  In
-	// any non-zero case, we should not try to do anything.
-	uint32_t expected = 0;
-	if (!barrier.compare_exchange_strong(expected, 1))
-	{
-		return false;
-	}
-	Debug::log("Initialising network interface");
-	auto &ethernet = lazy_network_interface();
-	ethernet.mac_address_set();
-	// Poke the barrier and make the driver thread start.
-	barrier = 2;
-	barrier.notify_one();
-	return true;
-}
 
 bool ethernet_send_frame(uint8_t *frame, size_t length)
 {
@@ -747,4 +749,46 @@ void firewall_remove_udpipv6_remote_endpoint(uint8_t *remoteAddress,
 		EndpointsTable<IPv6Address>::instance().remove_endpoint(
 		  IPProtocolNumber::UDP, *copy, localPort, remotePort);
 	}
+}
+
+bool ethernet_driver_start(std::atomic<uint8_t> *state)
+{
+	if (tcpipRestartState == nullptr)
+	{
+		if (!CHERI::check_pointer<CHERI::PermissionSet{
+		      CHERI::Permission::Load, CHERI::Permission::Global}>(
+		      state, sizeof(*state)))
+		{
+			Debug::log("Invalid TCP/IP state pointer {}", state);
+			return false;
+		}
+		tcpipRestartState = state;
+	}
+	if (tcpipRestartState->load() != 0)
+	{
+		// This is a restart, no need to actually reset the driver.
+		// Instead, just remove all firewall entries.
+		Debug::log("Network stack restart: clearing all entries.");
+		EndpointsTable<IPv6Address>::instance().clear(IPProtocolNumber::UDP);
+		EndpointsTable<IPv6Address>::instance().clear(IPProtocolNumber::TCP);
+		EndpointsTable<uint32_t>::instance().clear(IPProtocolNumber::UDP);
+		EndpointsTable<uint32_t>::instance().clear(IPProtocolNumber::TCP);
+		return true;
+	}
+	// Protect against double entry.  If the barrier state is 0, no
+	// initialisation has happened and we should proceed.  If it's 1, we're in
+	// the middle of initialisation, if it's 2 then initialisation is done.  In
+	// any non-zero case, we should not try to do anything.
+	uint32_t expected = 0;
+	if (!barrier.compare_exchange_strong(expected, 1))
+	{
+		return false;
+	}
+	Debug::log("Initialising network interface");
+	auto &ethernet = lazy_network_interface();
+	ethernet.mac_address_set();
+	// Poke the barrier and make the driver thread start.
+	barrier = 2;
+	barrier.notify_one();
+	return true;
 }
--- a/lib/tcpip/BufferManagement.cc
+++ b/lib/tcpip/BufferManagement.cc
@@ -3,6 +3,7 @@
 
 #include "FreeRTOS.h"
 #include "FreeRTOS_IP.h"
+#include <BufferManagement.hh>
 #include <NetworkBufferManagement.h>
 #include <cstdlib>
 #include <debug.hh>
@@ -11,13 +12,6 @@
 using Debug = ConditionalDebug<false, "Buffer management">;
 
 using CHERI::Capability;
-
-// Use a separate allocator quota for the buffer manager (false by default).
-// The buffer manager is responsible for allocating network buffers, which
-// differs from the other types of allocations the TCP/IP stack performs. It
-// may thus make sense to give it its own quota. We may want to expose this as
-// a build system configuration option at some point.
-#define USE_DEDICATED_BUFFERMANAGER_POOL false
 
 #if USE_DEDICATED_BUFFERMANAGER_POOL
 // Size of the buffer manager allocator quota.
@@ -29,6 +23,14 @@ DECLARE_AND_DEFINE_ALLOCATOR_CAPABILITY(BufferManagementMallocQuota,
                                         BM_MALLOC_QUOTA);
 #	define BM_MALLOC_CAPABILITY                                               \
 		STATIC_SEALED_VALUE(BufferManagementMallocQuota)
+
+/**
+ * Free the buffer manager's memory.
+ */
+void free_buffer_manager_memory()
+{
+	heap_free_all(BM_MALLOC_CAPABILITY);
+}
 #else
 #	define BM_MALLOC_CAPABILITY MALLOC_CAPABILITY
 #endif
@@ -132,7 +134,9 @@ void vReleaseNetworkBufferAndDescriptor(
 
 		// Failure is not supposed to happen unless we have a bug here
 		// or in FreeRTOS.
-		Debug::Assert(ret == 0, "Failed to free network buffer or descriptor.");
+		Debug::Assert(ret == 0,
+		              "Failed to free network buffer or descriptor (error {}).",
+		              ret);
 	}
 }
 
--- a/lib/tcpip/FreeRTOSIPConfig.h
+++ b/lib/tcpip/FreeRTOSIPConfig.h
@@ -33,6 +33,28 @@
 #define configMINIMAL_STACK_SIZE 128
 
 /**
+ * FreeRTOS+TCP uses descriptors to track information about the packets in a
+ * window. A pool of descriptors is allocated when the first TCP connection is
+ * made. This configuration entry sets the size of the pool.
+ *
+ * In the worst case, each TCP connection will have an Rx and Tx window of at
+ * most 8 segments (according to the FreeRTOS+TCP documentation), requiring 16
+ * descriptors in total per TCP connection. This configuration entry should
+ * thus be set to `(expected maximum number of TCP connections) x 16`.
+ *
+ * If we assume at most 6 TCP connections, this results in 96 segments.
+ *
+ * Each descriptor is 104 bytes, i.e., set to 96 this results in a memory
+ * footprint of 9,984 bytes.
+ *
+ * Note that the pool of descriptors is reallocated at every network stack
+ * reset. Setting this number to a large value (e.g., 256, the FreeRTOS+TCP
+ * default) is known to cause reset failures due to fragmentation preventing us
+ * from reallocating the pool.
+ */
+#define ipconfigTCP_WIN_SEG_COUNT 96
+
+/**
  * Enable counting semaphore functionality in the build, as this is necessary
  * to build FreeRTOS+TCP.
  */
@@ -62,7 +84,6 @@
 #define ipconfigREPLY_TO_INCOMING_PINGS 1
 
 #define ipconfigNUM_NETWORK_BUFFER_DESCRIPTORS 8
-
 
 #define ipconfigTCP_RX_BUFFER_LENGTH ( 1280 )
 #define ipconfigTCP_TX_BUFFER_LENGTH ( 1280 )
@@ -97,3 +118,10 @@
 
 #define ipconfigALLOW_SOCKET_SEND_WITHOUT_BIND 0
 #define ipconfigUSE_NETWORK_EVENT_HOOK 1
+
+// Necessary to have FreeRTOS+TCP invoke our own `ipFOREVER` function.  As of
+// FreeRTOS+TCP 3.1.0 this only enables external `ipFOREVER`. It would be nice
+// to upstream a patch to have this variable renamed into something specific to
+// `ipFOREVER` to ensure that we do not enable additional undesired options in
+// future releases.
+#define TEST
--- a/lib/tcpip/network_wrapper.cc
+++ b/lib/tcpip/network_wrapper.cc
@@ -4,9 +4,11 @@
 #include <FreeRTOS.h>
 #include <FreeRTOS_IP.h>
 #include <FreeRTOS_ND.h>
+#include <tcpip_error_handler.h>
 
-#include "../firewall/firewall.h"
+#include "../firewall/firewall.hh"
 #include "network-internal.h"
+#include "tcpip-internal.h"
 
 #include <FreeRTOS_IP_Private.h>
 #include <NetAPI.h>
@@ -20,8 +22,39 @@
 using Debug            = ConditionalDebug<false, "Network stack wrapper">;
 constexpr bool UseIPv6 = CHERIOT_RTOS_OPTION_IPv6;
 
-// IP thread global lock. See comment in FreeRTOS_IP_wrapper.c.
+// IP thread global lock. See comment in `FreeRTOS_IP_wrapper.c`.
 extern struct FlagLockState ipThreadLockState;
+
+/**
+ * This function controls the FreeRTOS+TCP network thread event loop.
+ *
+ * Returning 0 will cause the event loop to return, which we want to do when a
+ * reset of the network stack is triggered.
+ */
+extern "C" int ipFOREVER(void)
+{
+	// We must interrupt the loop if a reset is ongoing (`Restarting`).
+	// However, we must still allow the loop to run during the early stages
+	// of a reset (`IpThreadKicked`). Note that it is absolutely fine if
+	// the state changes after the check, as we will detect that change in
+	// the next loop iteration.
+	uint8_t state = restartState.load();
+	return (state == 0) || ((state & IpThreadKicked) != 0);
+}
+
+/**
+ * Current socket epoch. This is used to detect sockets that belong to a
+ * previous instance of the network stack.
+ *
+ * This should not be reset by the error handler and is reset-critical.
+ */
+std::atomic<uint32_t> currentSocketEpoch = 0;
+
+// Network stack reset globals. See comments in `tcpip-internal.h`.
+std::atomic<uint8_t>                      restartState    = 0;
+std::atomic<uint8_t>                      userThreadCount = 0;
+ds::linked_list::Sentinel<SocketRingLink> sealedSockets;
+FlagLockPriorityInherited                 sealedSocketsListLock;
 
 using CHERI::Capability;
 using CHERI::check_pointer;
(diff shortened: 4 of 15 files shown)


davidchisnall:
It looks as if our BearSSL submodule is pointing at a commit that's after the release branch.  I think we did that because there was a bug fix that we needed after the release, but apparently the server has changed its configuration to not permit clones of individual commits that are not branches / tags?

lib/tcpip/network_wrapper.cc
```diff
@@ -23,6 +25,46 @@ constexpr bool UseIPv6 = CHERIOT_RTOS_OPTION_IPv6;
 // IP thread global lock. See comment in FreeRTOS_IP_wrapper.c.
 extern struct FlagLockState ipThreadLockState;
 
+extern "C" int ipFOREVER(void)
+{
+	// We must interrupt the loop if a reset is ongoing (`Restarting`).
+	// However, we must still allow the loop to run during the early stages
+	// of a reset (`IpThreadKicked`).
+	uint32_t state = restartState.load();
+	return (state == 0) || ((state & IpThreadKicked) != 0);
```
davidchisnall, line 34: Maybe comment here that it's fine if the state changes after the test here because we'll be kicked again in the next loop iteration?

lib/tcpip/network_wrapper.cc
```diff
@@ -23,6 +25,46 @@ constexpr bool UseIPv6 = CHERIOT_RTOS_OPTION_IPv6;
...
+}
+
+/**
+ * State machine of the restart process. Used for synchronization across the
+ * TCP/IP stack and with the firewall.
```
davidchisnall, line 39: What values can this have?  Is it just a counter?

lib/tcpip/tcpip-internal.h
```diff
@@ -0,0 +1,107 @@
...
+ * Flags for the state of restart.
+ */
+enum [[clang::flag_enum]] RestartState{
+  NotRestarting  = 0,
+  Restarting     = 1,
+  IpThreadKicked = 2,
+  DriverKicked   = 4,
+};
+
+using ChunkFreeLink = ds::linked_list::cell::Pointer;
```
nwf, line 25:
```suggestion
using SocketRingLink = ds::linked_list::cell::Pointer;
```

lib/tcpip/tcpip_error_handler.h
```diff
@@ -0,0 +1,366 @@
...
+
+	// Set the currently restarting flag. This will do several things:
+	// 1. ensure that only one call to this error handler triggers a reset
+	// 2. ensure that no thread enters the compartment while we are
+	//    restarting
+	// 3. reset the network thread whenever it wakes up
+	uint32_t expected = NotRestarting;
+	if (!restartState.compare_exchange_strong(expected, Restarting))
+	{
+		if (isIpThread && ((restartState.load() & IpThreadKicked) != 0))
```
nwf, line 97: I think it's possible that `restartState` has been altered between the `compare_exchange_strong` and here, so `IpThreadKicked` may have been asserted and isn't now.  That's probably fine, but maybe this could be

```suggestion
		if (isIpThread && ((expected & IpThreadKicked) != 0))
```

hlef: Actually I think that if the flag was cleared, it's because the concurrent restart actually finished between the two loads. In that case we do not want to print the log message. Still, I think that this is an extremely unlikely thing to happen (if it can happen at all), and it's a good idea to get rid of that load as you suggest.

lib/firewall/firewall.cc
```diff
@@ -748,3 +753,45 @@ void firewall_remove_udpipv6_remote_endpoint(uint8_t *remoteAddress,
...
+	{
+		if (!CHERI::check_pointer<CHERI::PermissionSet{
+		      CHERI::Permission::Load, CHERI::Permission::Global}>(
+		      state, sizeof(*state)))
+		{
+			Debug::log("Invalid TCP/IP state pointer {}", state);
+			return false;
+		}
+		TCPIPRestartState = state;
+	}
```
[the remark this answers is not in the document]
hlef: You mean checking that `TCPIPRestartState == state`? We could, but I am not sure how to react to a mismatch. The value of this pointer should never change because it is a global. Thus any mismatch should either be a bug or a malicious attempt at corrupting the firewall. Do you think it is worth adding the branch just for a debug print?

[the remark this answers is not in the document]
nwf-msr: I wasn't sure how we should react on an `else` branch, either.  But if the `TCPIPRestartState` object is just something that needs to be shared between the firewall and the network compartment, could we statically arrange for that sharing rather than pass it across the compartment boundary?

That is, could we have one or the other export it and the other import it and rely on the `cheriot-audit` tool to make sure nobody else is trying to get their hands on it?  If so, I think we could do away with the parameter, and so this test, entirely?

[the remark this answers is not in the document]
davidchisnall: We don’t (yet) have a mechanism to export globals from one compartment to another.

lib/firewall/firewall.cc
```diff
@@ -748,3 +753,45 @@ void firewall_remove_udpipv6_remote_endpoint(uint8_t *remoteAddress,
...
+	if (TCPIPRestartState->load() != 0)
+	{
+		// This is a restart, no need to actually reset the driver.
+		// Instead, just remove all firewall entries.
+		Debug::log("Network stack restart: clearing all entries.");
+		EndpointsTable<IPv6Address>::instance().clear(IPProtocolNumber::UDP);
+		EndpointsTable<IPv6Address>::instance().clear(IPProtocolNumber::TCP);
```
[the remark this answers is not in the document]
hlef: Right - I cannot really fix that for now though because your PR which enables to compile out the IPv6 stuff has not been merged yet: https://github.com/CHERIoT-Platform/network-stack/pull/26

I suggest to leave it as it is for now and fix it when we rebase, either here (if you PR gets merged first), or in your PR (if this one gets merged first).

lib/firewall/firewall.h
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: MIT
 
 #include <compartment.h>
+#include <atomic>
```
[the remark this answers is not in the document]
hlef: Right, this is probably the best thing to do if we want to keep the atomic argument... especially since this is an internal header.

lib/firewall/firewall.cc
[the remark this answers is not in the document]
nwf-msr: In lieu of exporting globals, could we export a getter for the moment?

lib/firewall/firewall.h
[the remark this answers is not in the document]
hlef: I did that again in a separate commit.

lib/firewall/firewall.cc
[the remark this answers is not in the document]
hlef: Sure! We could also export a getter. I suspect that it would take more code size though?

lib/tcpip/network_wrapper.cc
hlef: True, it's technically just a counter because the bits should be sequentially set. First `NotRestarting`, then `Restarting`, then `Restarting | IpThreadKicked`, then `Restarting | IpThreadKicked | DriverKicked`, then `NotRestarting` again. Still, I quite like the bit enum as it is. Is there a reason to go for the counter instead that I'm not seeing?

One thing which can be done for sure is reducing the size of the atomic for 32 bit to 8 bit.

hlef: Actually the main reason why I did this is because I just took what we were doing here: https://github.com/CHERIoT-Platform/network-stack/blob/main/lib/tcpip/startup.cc#L40

Although clearly here we don't need the flag stuff because we have a strictly sequential process.

lib/tcpip/tcpip-internal.h
```diff
@@ -0,0 +1,110 @@
...
+ * by the firewall, and we trust the firewall not to call us at inappropriate
+ * times during a restart. The firewall has access to `restartState` and knows
+ * when not to call - if it does call at an inapproriate time during reset,
+ * this is a firewall bug. Thus, do not check `restartState` here.
+ */
+auto with_restarting_checks_driver(auto operation, auto errorValue)
+{
+	userThreadCount++;
+	auto ret = operation();
+	userThreadCount--;
```
davidchisnall, line 108: Possibly a comment saying that this decrement will happen in the error handler if it crashes?

lib/tcpip/network_wrapper.cc
```diff
@@ -82,15 +109,37 @@ namespace
 	 * it to `operation`.  If the unsealing fails, return `-EINVAL`.  This is a
 	 * helper function used for operations on sockets.
 	 */
-	int with_sealed_socket(auto operation, Sealed<SealedSocket> sealedSocket)
+	int with_sealed_socket(bool                 isCloseOperation,
```
davidchisnall, line 112: Since this is true in only one(?) call site, maybe move it last and give it a default `false` parameter value?

Also, since it's statically known (and since we're specialising this function for all call sites anyway), maybe make it a template parameter and make the `if` that uses it `constexpr` so that we definitely remove it during codegen?

lib/tcpip/tcpip-internal.h
```diff
@@ -0,0 +1,110 @@
...
+	{
+		return reinterpret_cast<struct SealedSocket *>(
+		  reinterpret_cast<uintptr_t>(c) - offsetof(struct SealedSocket, ring));
+	}
+};
+
+extern FlagLockPriorityInherited                 sealedSocketsListLock;
+extern ds::linked_list::Sentinel<SocketRingLink> sealedSockets;
+extern std::atomic<uint8_t>                      restartState;
+extern std::atomic<uint8_t>                      userThreadCount;
```
davidchisnall, line 71: Doc comments should probably be here rather than on the definitions.

lib/tcpip/tcpip_error_handler.h
```diff
@@ -0,0 +1,367 @@
...
+
+// Locks and futexes to release as part of the reset
+extern struct FlagLockState       ipThreadLockState;
+extern struct RecursiveMutexState __CriticalSectionFlagLock;
+extern struct RecursiveMutexState __SuspendFlagLock;
+extern QueueHandle_t              xNetworkEventQueue;
+
+// APIs to call as part of the reset
+extern void free_buffer_manager_memory();
+extern void network_restart();
```
davidchisnall, line 29: Doc comments.

lib/tcpip/tcpip_error_handler.h
```diff
@@ -0,0 +1,367 @@
...
+		  "User thread TCP/IP stack error handler called!");
+		userThreadCount--;
+	}
+	else
+	{
+		DebugErrorHandler::log(
+		  "Network thread TCP/IP stack error handler called!");
+	}
+
+	// Manually unlock the sealed sockets lock list if it was held.
```
davidchisnall, line 81: Under what situations can this happen?  My expectation is that it's only if it crashes while adding to the linked list.  Is the list state well defined in that case?  Are operations correctly serialised with compiler barriers to ensure that the list is at least traversable in one direction?

lib/tcpip/tcpip_error_handler.h
```diff
@@ -0,0 +1,367 @@
...
+			DebugErrorHandler::log("The network thread crashed while "
+			                       "restarting. This may be unrecoverable.");
+		}
+
+		// Another instance of the error handler is running, do not do
+		// anything.
+		return;
+	}
+
+	DebugErrorHandler::log("Reset-ing the network stack.");
```
davidchisnall, line 117: Resetting

lib/tcpip/tcpip_error_handler.h
```diff
@@ -0,0 +1,367 @@
...
+
+	// We need to acquire the sealed sockets lock because we do not want
+	// the sealed sockets list to be in an inconsistent state when we go
+	// over it.
+	//
+	// Waiting to acquire the lock is fine, as we know that any thread
+	// which holds the it will eventually release it, either 1) exiting the
+	// critical section, or 2) crashing into it, in which case we (the
+	// error handler) will manually unlock it (see manual unlock above).
```
davidchisnall, line 126: In theory, a crashing thread may corrupt the state of this lock, but that would require some memory-safety bug which, hopefully, can't happen.

lib/tcpip/tcpip_error_handler.h
```diff
@@ -0,0 +1,367 @@
...
+			if (Capability{s}.is_valid() &&
+			    Capability{s->xEventGroup}.is_valid())
+			{
+				DebugErrorHandler::log("Destroying event group {}.",
+				                       s->xEventGroup);
+				eventgroup_destroy_force(MALLOC_CAPABILITY, s->xEventGroup);
+			}
+			else
+			{
+				DebugErrorHandler::log("Ignoring corrupted socket {}.", s);
```
davidchisnall, line 169: Presumably memory for the event group will be cleaned up by the `heap_free_all` later, even if the pointer to it is corrupted?

lib/tcpip/tcpip_error_handler.h
```diff
@@ -0,0 +1,367 @@
...
+		// memory again before terminating.
+		free_compartment_memory();
+
+		Timeout t{1};
+		thread_sleep(&t);
+	}
+
+	// Wait for the IP thread to reset (unless this error handler is
+	// running from the IP thread).
+	if (isUserThread)
```
davidchisnall, line 221: I don't really like the amount of work that we do here.  I'd much rather that we did all of the cleanup on the TCP/IP thread, which has a known and predictable amount of stack space.  Running an error handler on a user thread means that the amount of stack space is unpredictable and may be very small, so we run a significant risk of double fault.

Ideally, all that this error handler would do would be to set a flag that we check in `ipFOREVER` and on new socket calls and then force unwind.  All other cleanup would run on the TCP/IP worker thread which, if it works once, we know definitely has enough stack space for all of the cleanup.

lib/tcpip/tcpip_error_handler.h
```diff
@@ -0,0 +1,367 @@
...
+
+	// Free heap memory.  We must do this *again*, because threads may have
+	// allocated memory since the previous calls to `heap_free_all`.
+	DebugErrorHandler::log("Freeing heap memory.");
+	free_compartment_memory();
+
+	// Update the socket epoch. We want to do this after all threads have
+	// terminated in case some threads were allocating new sockets during
+	// the restart.
+	currentSocketEpoch++;
```
davidchisnall, line 246: If we do this at the start and end of the epoch, threads can simply fail if the epoch number is odd (or if it changes during the socket creation).

lib/tcpip/tcpip_error_handler.h
```diff
@@ -0,0 +1,367 @@
...
+ * extracted through a manual study of the compartment's code-base: this may
+ * therefore break if new releases of FreeRTOS+TCP introduce new locks. In the
+ * future, we may want to come up with a more systematic approach.
+ *
+ * This function is designed to be robust against most types of compartment
+ * corruption, however we do assume that:
+ * - 'reset-critical' data has not been corrupted
+ * - the control-flow of threads in the compartment has not been altered
+ */
+extern "C" void reset_network_stack_state()
```
davidchisnall, line 64: This should probably also explain where / when this is called in the network reset phase.

lib/tcpip/tcpip_error_handler.h
```diff
@@ -0,0 +1,367 @@
...
+			    : *frame->get_register_value(registerNumber),
+			  threadID);
+
+			// TODO before running the reset function we should go
+			// to the top of the stack to ensure that we do not run
+			// out of stack space while executing the error
+			// handler.
+
+			// Reset the network stack state.
+			reset_network_stack_state();
```
davidchisnall, line 324: I think the correct thing to do is to make the IP thread call this function whenever it's doing a restart and have the code below set the PCC to a pointer to the run loop function with ca0 set to 1 telling it to do the reset.  Then the cleanup code always runs on a known thread.

lib/tcpip/tcpip_error_handler.h
hlef: Not necessarily. The thread can also crash in `network_socket_create_and_bind` while holding the lock (but not editing the list) because we hold the lock for a bit longer to simplify error handling.

We don't guarantee reset if the list is corrupted (it is "reset critical").

I added comments to clarify both of this.

lib/tcpip/BufferManagement.cc
```diff
@@ -33,6 +33,14 @@ DECLARE_AND_DEFINE_ALLOCATOR_CAPABILITY(BufferManagementMallocQuota,
 #	define BM_MALLOC_CAPABILITY MALLOC_CAPABILITY
 #endif
 
+void free_buffer_manager_memory()
+{
+#if USE_DEDICATED_BUFFERMANAGER_POOL
+	heap_free_all(BM_MALLOC_CAPABILITY);
+#endif
+	return;
+}
```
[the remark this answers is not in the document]
hlef: Good idea, done.

lib/tcpip/FreeRTOSIPConfig.h
```diff
@@ -58,3 +57,6 @@
 
 #define ipconfigALLOW_SOCKET_SEND_WITHOUT_BIND 0
 #define ipconfigUSE_NETWORK_EVENT_HOOK 1
+
+// Necessary to have FreeRTOS+TCP invoke our own `ipFOREVER` function.
+#define TEST
```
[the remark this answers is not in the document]
hlef: AFAIK this does not enable anything else. I added a comment, pushing that upstream would be nice indeed!

lib/tcpip/network_wrapper.cc
```diff
@@ -23,6 +25,48 @@ constexpr bool UseIPv6 = CHERIOT_RTOS_OPTION_IPv6;
 // IP thread global lock. See comment in FreeRTOS_IP_wrapper.c.
 extern struct FlagLockState ipThreadLockState;
 
+extern "C" int ipFOREVER(void)
```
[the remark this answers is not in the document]
hlef: Agree, done.

lib/tcpip/tcpip-internal.h
hlef: Yes! This is also where I want to say that things will go BAD if we don't run the error handler on stack overflow.

lib/tcpip/network_wrapper.cc
hlef: Good idea, thanks!

lib/tcpip/tcpip_error_handler.h
hlef: The bits about cleaning up the locks we cannot avoid because they are necessary to unstuck the IP thread :(

(apart from the socket locks)

The bits at the end (after the `userThreadCount.load() == 0`) could be moved elsewhere. As we discussed separately this would be necessary to be more resilient against the situation where a thread enters the compartment with a very small stack.

Unfortunately this is not trivial to do, because the current setup process of the TCP/IP stack is from two threads: it's the user thread which initializes the TCP/IP stack through `network_start`, not the IP thread. Then the user thread unblocks the IP thread through the atomic entry guard. We need to rework that logic to do the initialization from the IP thread.

As discussed separately, let's keep a TODO list in an issue to track this an move with the PR as it is.

hlef: To clarify this in the code, I added comments breaking down the code into bits which could be moved out of the error handler, and those which have to stay there.

lib/tcpip/tcpip_error_handler.h
hlef: I initially decided not to do it, mainly because we still need the `restartState` (which is currently used to detect a restart) because this is our state machine and it's more complex that just "is restart" or "is not restart".

As we discussed separately however, we could merge the `restartState` and the epoch into a single atomic.

We leave this for future works, I added a comment.

lib/tcpip/tcpip_error_handler.h
hlef: Correct, I do not know how attackers could corrupt the lock state without violating spatial and temporal memory safety. And if attackers manage that, it's something we need to fix in the locks I guess!

lib/tcpip/tcpip_error_handler.h
hlef: Right, the memory of the event group will still be freed later with the `heap_free_all`, however we run the risk to have the IP thread stuck on the event queue which we didn't manage to destroy.

hlef: I added a comment to clarify this.

lib/tcpip/tcpip_error_handler.h
hlef: As we discussed, it would indeed be nice to move as much as we can of the reset to the context of the IP thread, however some bits have to stay here to be able to unblock the IP thread in the first place. These bits may actually need quite a bit of stack as well. Thus, I still think that we should try to reset to the top of the stack if possible.

hlef:
@davidchisnall I have addressed all comments, and rebased the PR. This is ready for another round of review :slightly_smiling_face:

hlef:
> It looks as if our BearSSL submodule is pointing at a commit that's after the release branch. I think we did that because there was a bug fix that we needed after the release, but apparently the server has changed its configuration to not permit clones of individual commits that are not branches / tags?

For the record, this was addressed in https://github.com/CHERIoT-Platform/network-stack/pull/28

hlef:
As discussed with @davidchisnall, this looks fine to merge now. It's working well. It has known limitations, which are documented in the `README.md`. I will open an issue to keep track of these limitations, and ensure that we end up addressing them.

PR #28, opened 2024-06-27
Update BearSSL and FreeRTOS+TCP submodules

hlef:
- Bumping BearSSL should fix the issue we are currently encountering with CI being unable to clone the repository.

- Bumping FreeRTOS+TPC allows us to benefit from new security fixes, as well as from a fix to an issue which affected our network stack restart. See issue #1149 of the FreeRTOS-Plus-TCP repository: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/issues/1149

   This is a relatively big bump, so we need to check if everything is still working.


hlef:
OK, bumping FreeRTOS+TCP seems to break things, so I removed it from the PR to only bump BearSSL, which fixes the CI.

PR #29, opened 2024-06-28
Bump FreeRTOS+TCP submodule to 4.2.0

hlef:
Bump FreeRTOS+TCP submodule to 4.2.0.

This allows us to benefit from new security fixes, as well as from a fix to an issue which affected our network stack restart. See issue #1149 of the FreeRTOS-Plus-TCP repository.

This also comes with fixes need to compile and run with the new FreeRTOS+TCP.

### DIFF
--- a/lib/tcpip/FreeRTOSIPConfig.h
+++ b/lib/tcpip/FreeRTOSIPConfig.h
@@ -2,6 +2,42 @@
 // SPDX-License-Identifier: MIT
 
 #pragma once
+
+#ifndef FREERTOS_CONFIG_H
+/**
+ * This macro would usually be defined in FreeRTOSConfig.h, which we do not
+ * implement here because we do not use the FreeRTOS core. We define it here
+ * because FreeRTOS+TCP refuses to compile without.
+ */
+#define FREERTOS_CONFIG_H
+#endif
+
+/**
+ * FreeRTOS+TCP errno codes. See the header documentation.
+ */
+#include <FreeRTOS_errno.h>
+
+/**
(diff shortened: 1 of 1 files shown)


lib/tcpip/FreeRTOSIPConfig.h
```diff
@@ -2,6 +2,40 @@
 // SPDX-License-Identifier: MIT
 
 #pragma once
+
+/**
+ * This macro would usually be defined in FreeRTOSConfig.h, which we do not
+ * implement here because we do not use the FreeRTOS core. We define it here
+ * because FreeRTOS+TCP refuses to compile without.
+ */
+#define FREERTOS_CONFIG_H
```
davidchisnall, line 11:
```suggestion
#ifndef FREERTOS_CONFIG_H
#define FREERTOS_CONFIG_H
#endif
```

lib/tcpip/FreeRTOS_errno_TCP.h
```diff
@@ -0,0 +1,54 @@
...
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+/**
+ * This header defines errno values used by FreeRTOS+TCP.
+ *
+ * `FreeRTOS_errno_TCP.h` used to be part of the FreeRTOS+TCP tree, but was
+ * removed because integrated in the FreeRTOS core, which we do not use. We
+ * therefore re-introduce it here.
```
davidchisnall, line 11: We should probably add this in the FreeRTOS compat layer.  It would also be good to define these in terms of the ones in `errno.h`, rather than as constant literals.

PR #37, opened 2024-10-04
tls.cc precisely_bound_buffer representation fix

nwf:
@rmn30 points out that just because the requested length is shorter than the maximum representable length doesn't mean that it's representable. Round the requested length down by the alignment mask first!


davidchisnall:
Can you expand the comment? I’m not sure I understand the masking.

nwf:
Sure thing.  I'll copy commentary from https://github.com/CHERIoT-Platform/cheriot-sail/pull/74

nwf:
OBE; see #38.

PR #38, opened 2024-10-07
lib/tls: use RTOS bounds().set_inexact_up_to()

nwf:
https://github.com/CHERIoT-Platform/cheriot-rtos/pull/310 copied, fixed, and renamed our precisely_bound_buffer() into the RTOS Capability object, so use its spelling instead.

### DIFF
--- a/lib/tls/tls.cc
+++ b/lib/tls/tls.cc
@@ -189,56 +189,13 @@ namespace
 		br_ssl_engine_set_default_aes_gcm(&cc->eng);
 	}
 
-	/**
-	 * Set the bounds on `buffer` to `length` if `length` is representable with
-	 * the current alignment of `buffer`. If not, then reduce `length` until it
-	 * is representable.  This ensures that we call the TCP/IP stack with a
-	 * precisely bounded buffer, rather than one that is slightly larger,
-	 * avoiding potential corruption or information leaks if the TCP/IP stack
-	 * is compromised, at the expense of more cross-compartment calls.
-	 */
-	void precisely_bound_buffer(Capability<uint8_t> &buffer, size_t &length)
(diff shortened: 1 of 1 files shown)


lib/tls/tls.cc
```diff
@@ -275,7 +231,7 @@ namespace
 		size_t     readyLength;
 		Capability readyBuffer =
 		  br_ssl_engine_sendrec_buf(engine, &readyLength);
-		precisely_bound_buffer(readyBuffer, readyLength);
+		readyBuffer.bounds().set_inexact_at_most(readyLength);
```
davidchisnall, line 234: I think line 241 now needs to be .length of this? Otherwise, I would expect a bounds failure.

nwf: `set_inexact_at_most` takes its argument by reference and updates it, so the length of `readBuffer`'s bounds should equal `readyLength`?  Am I missing something?

davidchisnall: Oh, hmm, I missed that. It is inconsistent with the other CHERI::Capability APIs, so we should probably change it or rename the API to make it explicit.

nwf: Sorry, I just stole the API from here.  Open to suggestions for amendments.

I guess when we do switch to `CSetBoundsRoundDown`, we'll have to follow it with the `CGetLength` that this write-back elides.  Still, probably pairing those two in this function is a fine thing to do.

davidchisnall: I would rather remove the writeback. The get length is an explicit operation and the use is easier to follow. I *think* the compiler can elide an unused bounds-get intrinsic, but I’m not 100% sure.

PR #39, opened 2024-10-09
Add a server API.

hlef:
This PR implements a server API, and showcases it with a toy HTTP server.

This has been tested in normal conditions, through various edge-cases (retransmitted SYNs, incomplete TCP create three-way handshake, client-side termination and server-side termination), as well as through various network stack resets. I also tested it for memory leaks.

### DIFF
--- a/examples/05.HTTP_SERVER/http_server.cc
+++ b/examples/05.HTTP_SERVER/http_server.cc
@@ -1,0 +1,218 @@
+#include "timeout.h"
+#include <NetAPI.h>
+#include <debug.hh>
+#include <errno.h>
+#include <fail-simulator-on-error.h>
+#include <memory>
+#include <string_view>
+#include <thread.h>
+#include <tick_macros.h>
+
+using CHERI::Capability;
+
+using Debug            = ConditionalDebug<true, "HTTP server example test">;
+constexpr bool UseIPv6 = CHERIOT_RTOS_OPTION_IPv6;
+
+/**
+ * Bind capability for the server port 80. Use IPv6 if enabled in the
+ * configuration, and allow at most two simultaneous connections to the server.
+ */
+#define LISTEN_PORT 80
+DECLARE_AND_DEFINE_BIND_CAPABILITY(HTTPPort, UseIPv6, LISTEN_PORT, 10);
+
+DECLARE_AND_DEFINE_ALLOCATOR_CAPABILITY(TestMalloc, 32 * 1024);
+#define TEST_MALLOC STATIC_SEALED_VALUE(TestMalloc)
+
+static char reply[] =
+  "HTTP/1.1 200 OK\r\n"
+  "Content-type: text/html\r\n"
+  "Connection: close\r\n"
+  "\r\n"
+  "<!DOCTYPE HTML PUBLIC \"-//IETF//DTD HTML 2.0//EN\">"
+  "<html>"
+  "<head><title>Hello from CHERIoT!</title></head>"
+  "<body><h1>It works!</h1><p>Served from a CHERIoT device.</p></body>"
+  "</html>\n";
+
+/**
+ * Maximum number of clients the server will server before shutting down. This
+ * is useful to check that the server can handle multiple clients before
+ * terminating.
+ */
+static const uint16_t MaxClients = 10;
+
+/**
+ * This example server is written to showcase the network stack server API, but
+ * also the network stack restart. If one of the network stack APIs fails in a
+ * way likely caused by a crash, the server will wait this small delay before
+ * retrying to give time to the network stack to reset.
+ */
+static const uint16_t RestartDelay = 100; // in ticks
+
+void __cheri_compartment("http_server_example") example()
+{
+	network_start();
+
+	auto heapAtStart = heap_quota_remaining(TEST_MALLOC);
+
+	uint16_t clientsCounter = 0;
+
+	Debug::log("Starting the server.");
+
+	while (clientsCounter < MaxClients)
+	{
+		Debug::log("Creating a listening socket.");
+		Timeout unlimited{UnlimitedTimeout};
+		auto    socket = network_socket_listen_tcp(
+		     &unlimited, TEST_MALLOC, STATIC_SEALED_VALUE(HTTPPort));
+
+		if (!Capability{socket}.is_valid())
+		{
+			Debug::log("Failed to create a listening socket.");
+			// This may have failed because of a network stack
+			// crash. Sleep a little bit to enable a reset.
+			Timeout sleep{RestartDelay};
+			thread_sleep(&sleep);
+			continue;
+		}
+
+		Debug::log("Listening on port {}...", LISTEN_PORT);
+		while (clientsCounter < MaxClients)
+		{
+			NetworkAddress clientAddress = {0};
+			uint16_t       clientPort    = 0;
+
+			auto clientSocket = network_socket_accept_tcp(
+			  &unlimited, TEST_MALLOC, socket, &clientAddress, &clientPort);
+
+			if (!Capability{clientSocket}.is_valid())
+			{
+				Debug::log("Failed to establish a connection.");
+				Timeout sleep{RestartDelay};
+				thread_sleep(&sleep);
+				break;
+			}
+
+			if (!UseIPv6)
+			{
+				Debug::log("Established a connection with {}.{}.{}.{}:{}",
+				           static_cast<int>(clientAddress.ipv4) & 0xff,
+				           static_cast<int>(clientAddress.ipv4 >> 8) & 0xff,
+				           static_cast<int>(clientAddress.ipv4 >> 16) & 0xff,
+				           static_cast<int>(clientAddress.ipv4 >> 24) & 0xff,
+				           clientPort);
+			}
+			else
+			{
+				Debug::log("Established a connection.");
+			}
+
+			clientsCounter++;
+
+			auto [received, buffer] =
+			  network_socket_receive(&unlimited, TEST_MALLOC, clientSocket);
+
+			// For this simple server, we do not care about what the client
+			// sent (we will always serve the same content).
+			int ret = heap_free(TEST_MALLOC, buffer);
+			if (ret != 0)
+			{
+				// This may happen if the network stack
+				// crashed: if the network stack crashes during
+				// `network_socket_receive` it is likely that
+				// `buffer` will be a nullptr or any other
+				// untagged value, and `heap_free` will return
+				// an error when passed the invalid thing.
+				Debug::log("Failed to free receive buffer: {}", ret);
+				// Do not break here - we still want to free the socket.
+			}
+
+			if (received > 0)
+			{
+				Debug::log(
+				  "Received {} bytes from the client, serving static content.",
+				  received);
+				constexpr size_t ToSend = sizeof(reply) - 1;
+				size_t           sent   = 0;
+				while (sent < ToSend)
+				{
+					size_t remaining = ToSend - sent;
+
+					ssize_t sentThisCall = network_socket_send(
+					  &unlimited, clientSocket, &(reply[sent]), remaining);
+					Debug::log("Sent {} bytes", sentThisCall);
+
+					if (sentThisCall >= 0)
+					{
+						sent += sentThisCall;
+					}
+					else
+					{
+						Debug::log("Send failed: {}", sentThisCall);
+						break;
+					}
+				}
+			}
+			else
+			{
+				Debug::log(
+				  "Failed to receive request from the client, error {}.",
+				  received);
+			}
+
+			Debug::log("Terminating the connection with the client.");
+			int retries = 10;
+			// In a retry loop to be more rebust to network stack crashes.
+			for (; retries > 0; retries--)
+			{
+				if (network_socket_close(
+				      &unlimited, TEST_MALLOC, clientSocket) == 0)
+				{
+					break;
+				}
+				Timeout sleep{RestartDelay};
+				thread_sleep(&sleep);
+			}
+
+			if (retries == 0)
+			{
+				Debug::log("Failed to close the client socket.");
+			}
+		}
+
+		Debug::log("Closing the listening socket.");
+		int retries = 10;
+		// In a retry loop to be more rebust to network stack crashes.
+		for (; retries > 0; retries--)
+		{
+			if (network_socket_close(&unlimited, TEST_MALLOC, socket) == 0)
+			{
+				break;
+			}
+			Timeout sleep{RestartDelay};
+			thread_sleep(&sleep);
+		}
+
+		if (retries == 0)
+		{
+			Debug::log("Failed to close the listening socket.");
+		}
+	}
+
+	Debug::log("Now checking for leaks.");
+	auto heapAtEnd = heap_quota_remaining(TEST_MALLOC);
+	if (heapAtEnd < heapAtStart)
+	{
+		Debug::log("Warning: The implementation leaked {} bytes (start: {} vs. "
+		           "end: {}).",
+		           heapAtStart - heapAtEnd,
+		           heapAtEnd,
+		           heapAtStart);
+	}
+	else
+	{
+		Debug::log("No leaks detected.");
+	}
+
+	Debug::log("Terminating the server.");
+}
--- a/examples/05.HTTP_SERVER/xmake.lua
+++ b/examples/05.HTTP_SERVER/xmake.lua
@@ -1,0 +1,62 @@
+-- Copyright CHERIoT Contributors.
+-- SPDX-License-Identifier: MIT
+
+-- Update this to point to the location of the CHERIoT SDK
+sdkdir = path.absolute("../../../cheriot-rtos/sdk")
+
+set_project("CHERIoT HTTP Server Example")
+
+includes(sdkdir)
+
+set_toolchains("cheriot-clang")
+
+includes(path.join(sdkdir, "lib"))
+includes("../../lib")
+
+option("board")
+  set_default("ibex-arty-a7-100")
+
+compartment("http_server_example")
+  set_default(false)
+  add_includedirs("../../include")
+  add_deps("freestanding", "TCPIP", "NetAPI")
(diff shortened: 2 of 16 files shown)


hlef:
For testing, I found hping3 particularly useful:
- `hping3 -c 1 -S -p 80 $ip` for sending a single SYN (incomplete three-way handshake)
- `hping3 -c 1 -S -p 80 $ip -s 9999` for sending a single SYN that reuses the same client port 9999 (simulate SYN retransmissions)
- `hping3 -c 1 -1 $ip` for sending a ping of the death (see patch [here](https://github.com/CHERIoT-Platform/network-stack/pull/27))

I simply used curl to grab the webpage this server serves: `curl $ip`.

It is worth noting that this server does not properly close the TCP connection with the client, and so curl has to be forcefully terminated (e.g., with `CTRL-C`). This is due to a pre-existing bug which I documented in #40 .

Also, the server seems to indicate a memory leak when a network stack reset occurs. I suspect that this is also a pre-existing problem which should be investigated separately.

lib/tcpip/network_wrapper.cc
```diff
@@ -590,6 +689,128 @@ SObj network_socket_create_and_bind(Timeout       *timeout,
...
+			  // This cannot fail unless buggy - we know that we
+			  // successfully allocated the token with this malloc
+			  // capability. Same for other calls to `token_obj_destroy`
+			  // in this function.
+			  token_obj_destroy(mallocCapability, socket_key(), sealedSocket);
+			  return -EINVAL;
+		  }
+		  socketWrapper->socket = rawSocket;
+
+		  // Claim the socket so that it counts towards the caller's quota.  The
```
davidchisnall, line 733: I think this is copied from my code, but I think it's wrong and should be fixed in both places.

If we allocate with the network stack's capability and claim with the caller's capability, then we are the owner.  We cannot free with heap_free_all because it needs the sealing key, but they can call heap_free_all to drop their claim.  You can use this to exhaust the network stack's quota.

lib/tcpip/network_wrapper.cc
```diff
@@ -590,6 +689,128 @@ SObj network_socket_create_and_bind(Timeout       *timeout,
...
+		  {
+			  // See above comments.
+			  Debug::log("Failed to add socket to the socket reset list.");
+			  close_socket_retry(timeout, rawSocket);
+			  token_obj_destroy(mallocCapability, socket_key(), sealedSocket);
+			  return -EINVAL;
+		  }
+
+		  // Set `address`.
+		  if ((heap_claim_fast(timeout, address) < 0) ||
```
davidchisnall, line 768: You shouldn't need to claim a timeout, since check_timeout checks that it's on the stack.

lib/tcpip/network_wrapper.cc
```diff
@@ -590,6 +689,128 @@ SObj network_socket_create_and_bind(Timeout       *timeout,
...
+				    address->ipv6, addressTmp.sin_address.xIP_IPv6.ucBytes, 16);
+			  }
+			  else
+			  {
+				  address->kind = NetworkAddress::AddressKindIPv4;
+				  address->ipv4 = addressTmp.sin_address.ulIP_IPv4;
+			  }
+		  }
+		  // Set `port`.
+		  if ((heap_claim_fast(timeout, port) < 0) ||
```
davidchisnall, line 795: As above.

examples/05.HTTP_SERVER/http_server.cc
```diff
@@ -0,0 +1,209 @@
...
+
+			auto [received, buffer] =
+			  network_socket_receive(&unlimited, TEST_MALLOC, clientSocket);
+
+			// For this simple server, we do not care about what the client
+			// sent (we will always serve the same content).
+			int ret = heap_free(TEST_MALLOC, buffer);
+			if (ret != 0)
+			{
+				// This may happen if the network stack crashed.
```
nwf, line 116: To check my understanding: because it may have been freed out from under us, and we aren't holding a claim on it?  That is, this isn't a memory leak.

examples/05.HTTP_SERVER/http_server.cc
```diff
@@ -0,0 +1,209 @@
...
+				Debug::log(
+				  "Received {} bytes from the client, serving static content.",
+				  received);
+				constexpr size_t toSend = sizeof(reply) - 1;
+				size_t           sent   = 0;
+				while (sent < toSend)
+				{
+					size_t remaining = toSend - sent;
+
+					size_t sentThisCall = network_socket_send(
```
nwf, line 132: Shouldn't that be `ssize_t` if we're returning negatives?

include/NetAPI.h
```diff
@@ -288,6 +325,23 @@ struct ConnectionCapability
...
+/**
+ * Bind capability contents. Instances of this sealed with the NetworkBindKey
+ * sealing capability authorise binding to a specific server port.
+ */
+struct BindCapability
+{
+	/**
+	 * Allow to bind on an IPv6 or IPv4 interface.
+	 */
+	bool isIPv6;
```
nwf, line 337: Maybe a pair of `PF_*` identifier and protocol index rather than just "{v4, v6}"?

lib/firewall/firewall.cc
```diff
@@ -586,6 +586,55 @@ namespace
...
+		uint16_t windowSize;
+		/**
+		 * Checksum.
+		 */
+		uint16_t checksum;
+		/**
+		 * Urgent pointer.
+		 */
+		uint16_t urgentPointer;
+	} __packed;
```
nwf, line 623: Just for paranoia, a `_Static_assert(sizeof(TCPHeader) == 20)`, perhaps.  (And maybe move the one for the `IPv4Header` below up above `TCPHeader`?)

include/NetAPI.h
hlef: Do you mean a `uint8_t domain` field with `AF_INET` and `AF_INET6` equivalents? Why not. That being said it complicates the API and I'm not sure that any other potential value for this field would make sense

examples/05.HTTP_SERVER/http_server.cc
```diff
@@ -0,0 +1,209 @@
...
+#include <thread.h>
+#include <tick_macros.h>
+
+using CHERI::Capability;
+
+using Debug            = ConditionalDebug<true, "HTTP server example test">;
+constexpr bool UseIPv6 = CHERIOT_RTOS_OPTION_IPv6;
+
+#define LISTEN_PORT 80
+DECLARE_AND_DEFINE_BIND_CAPABILITY(HTTPPort, UseIPv6, LISTEN_PORT);
```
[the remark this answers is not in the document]
hlef: As part of an separate discussion David and I decided that the current situation is actually OK: The capability currently is either for v4 or v6, depending on the value of `CHERIOT_RTOS_OPTION_IPv6`. The idea is that if a client needs to listen on both it would create two sockets with two different bind capabilities. But since this is just an example I took `CHERIOT_RTOS_OPTION_IPv6` to make the cut whether or not to use a v6 or v4 socket.

include/NetAPI.h
```diff
@@ -60,6 +60,43 @@ SObj __cheri_compartment("NetAPI")
...
+/**
+ * Create a listening TCP socket bound to a given port.
+ *
+ * The `mallocCapability` argument is used to allocate memory for the socket
+ * and must have sufficient quota remaining for the socket.
+ *
+ * The `bindCapability` argument is a capability authorising the bind to a
+ * specific server port.
+ *
+ * This returns a valid sealed capability to a socket on success, or a null on
```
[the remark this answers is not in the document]
hlef: Sounds good, agree. This is copy-pasted from other APIs we already had, so I would update there as well.

lib/firewall/firewall.hh
```diff
@@ -91,9 +102,20 @@ void __cheri_compartment("Firewall")
...
+ *
+ * `localPort` must not be a server port, as server port can be connected to
+ * many remote endpoints.
  */
 void __cheri_compartment("Firewall")
-  firewall_remove_tcpipv4_endpoint(uint16_t localPort);
+  firewall_remove_tcpipv4_local_endpoint(uint16_t localPort);
+
+/**
+ * Remove a specific remote TCP endpoint from the firewall.
```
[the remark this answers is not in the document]
hlef: Good point. These would be called by the network stack compartment when connections are teared down. I will add this to the documentation and network_stack.rego.

lib/tcpip/network_wrapper.cc
```diff
@@ -574,6 +645,34 @@ SObj network_socket_create_and_bind(Timeout       *timeout,
 				    mallocCapability, socket_key(), sealedSocket);
 				  return nullptr;
 			  }
+
+			  if (isListening)
+			  {
+				  auto listenResult =
+				    FreeRTOS_listen(socket, MAX_SIMULTANEOUS_TCP_CONNECTIONS);
```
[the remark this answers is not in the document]
hlef: Sounds good, seems that there could be legitimate use-cases for this.

lib/tcpip/network_wrapper.cc
hlef: David and I discussed this offline, and it sounds like this is a pre-existing, larger problem. I described the issue in https://github.com/CHERIoT-Platform/cheriot-rtos/issues/140 and we will address this separately.

lib/tcpip/network_wrapper.cc
hlef: I am confused - this code is doing a claim fast on the `address`, not the timeout?

lib/tcpip/network_wrapper.cc
hlef: Same here? I intended to claim fast the `port`.

lib/netapi/NetAPI.cc
```diff
@@ -120,17 +128,73 @@ SObj network_socket_connect_tcp(Timeout *timeout,
...
+	{
+		Debug::log("Failed to unseal bind capability");
+		return nullptr;
+	}
+
+	if constexpr (!UseIPv6)
+	{
+		if (serverPort->isIPv6)
+		{
+			Debug::log("IPv6 is not supported");
```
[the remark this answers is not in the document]
hlef: The way I thought `isIPv6` is to indicate whether the bind capability is for IPv4 exclusive or IPv6. I need to document that better!

examples/05.HTTP_SERVER/http_server.cc
hlef: Actually this may simply happen because the TCP/IP stack crashed during `network_socket_receive`, resulting in `buffer` being a nullptr, or... any other untagged thing. `heap_free` will then return an error when we pass it this invalid thing.

Thank you for this comment, this is worth documenting!

examples/05.HTTP_SERVER/http_server.cc
hlef: Agree!

include/NetAPI.h
nwf: I mean something like `struct BindCapability { uint8_t protocol_family; uint8_t protocol; uint16_t port }` so that I can capture at least `{IPv4, IPv6} x {TCP, UDP}`.  (We don't have to reuse `PF_*`/`AF_*` from Berkeley and we don't have to reuse their protocol namespaces within those families, but we could if convenient.)

davidchisnall: We treat TCP and UDP differently here (bind + accept doesn’t make sense for UDP).

I think I’d like a single capability to authorise v4 and v6, but what I really want probably isn’t possible: a single handle for a listening socket that lets you accept connections on either v4 or v6 if you’re supporting both.

The v6 case is more complex than v4 because you always have multiple v6 interfaces (ant least an link-local and routable one) and need to choose the one to listen on, so we might also need to expose that.

davidchisnall:
Just checking, but I think TLS is out of scope for this?  Connecting up the server bits of BearSSL is future work?

include/NetAPI.h
hlef: I think that we concluded that, for starters, we'd assume a single interface and evolve bind capabilities as needs arise. Shall we stay with how things are for now?

hlef:
> Just checking, but I think TLS is out of scope for this? Connecting up the server bits of BearSSL is future work?

Right, I would consider this future work. Hopefully that makes sense?

include/NetAPI.h
```diff
@@ -60,6 +60,43 @@ SObj __cheri_compartment("NetAPI")
...
+SObj __cheri_compartment("NetAPI")
+  network_socket_listen_tcp(Timeout *timeout,
+                            SObj     mallocCapability,
+                            SObj     bindCapability);
+
+/**
+ * Accept a connection on a listening socket.
+ *
+ * This function will block until a connection is established or the timeout is
+ * reached.
```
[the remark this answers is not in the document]
hlef: To expand a bit on this, the idea is that we also need this API to integrate in the [multiwait](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/main/sdk/core/scheduler/multiwait.h) API, which allows a waiter to wait on multiple events. And to do that, we need to expose a futex.

Does that make sense @davidchisnall ?

lib/tcpip/network_wrapper.cc
```diff
@@ -484,11 +491,75 @@ int network_host_resolve(const char     *hostname,
 	  -1 /* invalid if we are restarting */);
 }
 
+/**
+ * Callback called by FreeRTOS+TCP when a TCP connection is created or
+ * terminated.
+ *
+ * We use this callback to handle the case where a three-way TCP handshake
+ * initiated by a peer on a listening socket fails.
```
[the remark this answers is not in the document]
hlef: Assuming this gets called when a socket is ready to be accepted (I think it should but I need to run it and check), this indeed sounds like a good place to plug the futex in.

include/NetAPI.h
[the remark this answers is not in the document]
davidchisnall: Yes, exactly. Think of the multiwaiter as the equivalent of poll or select. Anything that you might want to use with these APIs should expose a futex.

include/NetAPI.h
davidchisnall: Probably fine for now, but maybe add a comment that this API is less stable than the rest of the stack and will change?

We also need corresponding rego functions to decode these capabilities.

davidchisnall:
(Rebasing to pick up CI fixes)

davidchisnall:
Now the CI failures are new, not ones that were present prior to this PR.

hlef:
> Now the CI failures are new, not ones that were present prior to this PR.

Thanks for rebasing! I'll fix that as well.

examples/05.HTTP_SERVER/xmake.lua
```diff
@@ -0,0 +1,62 @@
...
+set_toolchains("cheriot-clang")
+
+includes(path.join(sdkdir, "lib"))
+includes("../../lib")
+
+option("board")
+  set_default("ibex-arty-a7-100")
+
+compartment("http_server_example")
+  set_default(false)
```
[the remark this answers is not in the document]
hlef: I do not actually understand what the `set_default` does. I seem to understand that when passed `false` the compartment will not be compiled unless used (=depended on by something compiled)?

include/NetAPI.h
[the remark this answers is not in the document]
hlef: As we discussed offline, this is actually something that we should push consistently across the network stack API. This is a fair amount of work, so I opened an issue (https://github.com/CHERIoT-Platform/network-stack/issues/44) to document and suggest that we do this as part of a separate PR.

lib/tcpip/network_wrapper.cc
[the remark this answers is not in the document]
hlef: As discussed earlier, this is probably something we want to do as a separate PR, but so that we don't forget it I added a note to https://github.com/CHERIoT-Platform/network-stack/issues/44.

examples/05.HTTP_SERVER/xmake.lua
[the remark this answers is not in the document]
davidchisnall: Yes, but since this is a compartment that that is a direct dependency of the root target, it is not necessary. Use false when you are creating a library and want people to be able to include a load of libraries in xmake but not build them unless they’re used.

examples/05.HTTP_SERVER/http_server.cc
hlef: Actually, the code of other examples was wrong, so I fixed it there as well.

lib/tcpip/network_wrapper.cc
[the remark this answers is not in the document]
hlef: More thoughts about "Maybe put that in the bind-and-listen capability?":

I view the purpose of bind and connection capabilities as a way to enforce least privilege at the network stack API.

Thus, I need to understand which privilege we would be enforcing by putting the "maximum number of TCP connections that can be created (by remote peers) into a given server port" into the bind capability.

Is the idea to limit how many TCP connections a compartment can create in total? I can indeed see a problem where compartment A prevents compartment B from establishing network connections, by using up all the TCP connection slots in the network stack (which is quite limited in practice, typically <10). But that problem already exists pretty badly in the network stack, since connection capabilities do not come with such a limit: they can be used as many times as a compartment wants to open as many connections as they want.

This makes me think that this is a broader issue in the API, and that we may want to have a fix directly dedicated at this problem that fixes both bind and connection capabilities together. It may not be trivial to fix to match real-world use-cases... I don't know if the limit should be global for one compartment, or local to each connection / bind capability. The latter may be difficult to use in practice.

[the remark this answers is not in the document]
hlef: As we discussed online, let's put that maximum value in the bind capability and solve the problem separately for the connection capability: https://github.com/CHERIoT-Platform/network-stack/issues/45

hlef:
@davidchisnall @nwf All comments addressed! I re-tested with the new changes. I also ensured that the rego audit changes pass as expected (and needed a few fixes to pre-existing issues for that).

davidchisnall (approved):

nwf (approved):

Issue #43, opened 2024-10-16
MQTT ring buffer should be protected in subscribe callbacks 


PhilDay-CT:
The MQTT subscribe callback gets passed a reference to part of the receive buffer used by coreMQTT for incoming messages.  We should pass this as a read-only, non-capturable capabilty, since outside of the context of the callback it's content is not valid.

hlef:
Oops, sounds right, looks like I missed this. I added it to my TODO list, but go ahead if you want to do the fix!

hlef:
Note: I fixed this in my `hlefeuvre/bug-fixes` branch, which I will PR once the server API has been merged.

hlef:
This was fixed as part of #46.

PR #46, opened 2024-10-24
Various bug fixes in the network stack

hlef:
This PR clears some of my bugfix TODO list. This addresses a lot of documentation issues, but also actual bugs such as https://github.com/CHERIoT-Platform/network-stack/issues/43.

### DIFF
--- a/lib/tcpip/network-internal.h
+++ b/lib/tcpip/network-internal.h
@@ -46,7 +46,7 @@ __cheri_compartment("TCPIP") int network_host_resolve(
  *
  * This returns a sealed capability to a socket on success, or null on failure.
  *
- * This should be called only from the NetAPI compartment.
+ * This should be called only from the NetAPI or TCP/IP compartments.
  */
 SObj __cheri_compartment("TCPIP")
   network_socket_create_and_bind(Timeout       *timeout,
--- a/include/NetAPI.h
+++ b/include/NetAPI.h
@@ -219,8 +219,8 @@ int __cheri_compartment("TCPIP")
  * freeing this buffer.
  *
  * The `address` and `port` arguments are used to return the address and port
- * of the sender.  These can be null if the caller is not interested in the
- * sender's address or port.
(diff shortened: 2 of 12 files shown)


include/endianness.h
```diff
@@ -0,0 +1,26 @@
+// Copyright SCI Semiconductor and CHERIoT Contributors.
+// SPDX-License-Identifier: MIT
```
davidchisnall, line 2: Since this is C++-only, it should probably be named .hh.

lib/tcpip/network-internal.h
```diff
@@ -46,7 +46,7 @@ __cheri_compartment("TCPIP") int network_host_resolve(
  *
  * This returns a sealed capability to a socket on success, or null on failure.
  *
- * This should be called only from the NetAPI compartment.
+ * This should be called only from the NetAPI or TCP/IP compartments.
```
davidchisnall, line 49: Why is it called from the TCP/IP compartment? Wouldn’t that use internal APIs?

This contradicts the property enforced in rego:
https://github.com/CHERIoT-Platform/network-stack/blob/473249a6ace45f1d1fe9a1606705355b190327b3/network_stack.rego#L167

hlef: We use it in `network_socket_udp`, which is just a light wrapper for `network_socket_create_and_bind` with connection type UDP. I'm not sure using an internal API is worth it, the current code makes sense. We should update the rego as well, will update.

hlef:
Sounds like CI is failing because of https://github.com/CHERIoT-Platform/cheriot-rtos/commit/c3c0fb1d4b2d865cb51081c952325f556e25a3f1

hlef:
Comments addressed!

hlef:
> Sounds like CI is failing because of [CHERIoT-Platform/cheriot-rtos@c3c0fb1](https://github.com/CHERIoT-Platform/cheriot-rtos/commit/c3c0fb1d4b2d865cb51081c952325f556e25a3f1)

Should be fixed now.

Issue #31, opened 2024-08-13
Network Stack Reset limitations


hlef:
We recently merged the network stack reset as part of https://github.com/CHERIoT-Platform/network-stack/pull/27.

Let us use this issue to keep track of the limitations of the reset:

- [ ] The TCP/IP stack **cannot currently recover from a crash due to a stack overflow in the TCP/IP compartment**.
       This is due to a limitation of the implementation of the switcher, which cannot trigger the error handler on stack overflow. We should make it possible for the error handler to run on stack overflow.
- [ ] A small set of **globals called *reset-critical* outlive resets and/or are necessary for the reset**.
      We assume that this data has not been corrupted. This data is correspondingly annotated in the source code. We should address this with an inner compartment.
- [ ] We assume that the **control-flow of threads in the compartment has not been altered**.
      Given that we have spatial and temporal memory safety, this should be a relatively sane assumption? We can revisit later - for now there is no plan to address this.
- [ ] Upon reset, **we do not close TCP connections**.
       This leads remote servers to send us retransmissions which, in the case of several consecutive resets, can result in flooding. Ideally we should enable the reset code to send TCP RST packets to close outstanding TCP connections during the network stack reset.

davidchisnall:
> The TCP/IP stack cannot currently recover from a crash due to a stack overflow in the TCP/IP compartment.

This can now be fixed.  CHERIoT-Platform/CHERIoT-RTOS#301 made it possible to run error handler code up the stack.

hlef:
> > The TCP/IP stack cannot currently recover from a crash due to a stack overflow in the TCP/IP compartment.
> 
> This can now be fixed. [CHERIoT-Platform/cheriot-rtos#301](https://github.com/CHERIoT-Platform/cheriot-rtos/pull/301) made it possible to run error handler code up the stack.

Done in #48!

PR #48, opened 2024-10-29
Move to a stack-overflow-resilient error handler

hlef:
The TCP/IP stack cannot currently recover from a crash due to a stack overflow in the TCP/IP compartment.  This is due to a limitation of the implementation of the switcher, which cannot trigger the error handler on stack overflow. Further, the current implementation of the error handler runs at the point in the stack where the compartment crashed. This implies that the error handler runs on a potentially very small stack, making it possible that it will itself crash due to a stack overflow. This limitation is documented in https://github.com/CHERIoT-Platform/network-stack/issues/31.

This commit addresses these limitations using David's recently-merged stackless error handler mechanism: https://github.com/CHERIoT-Platform/cheriot-rtos/pull/301.

We also address a bug in the reset, which happens when an error handler on a user thread triggers a crash of the IP thread. In this case, previously improperly tested, we fail to release a lock, resulting in a failure of the reset.

More information in each commit.

### DIFF
--- a/lib/tcpip/tcpip-internal.h
+++ b/lib/tcpip/tcpip-internal.h
@@ -5,12 +5,15 @@
 #include <FreeRTOS_IP.h>
 #include <ds/linked_list.h>
 #include <locks.hh>
+#include <unwind.h>
 
 /**
  * Internal helpers and data structures for use inside of the TCP/IP
  * compartment. These should be called or used only from/in the TCP/IP
  * compartment.
  */
+
+extern "C" void reset_network_stack_state(bool isIpThread);
 
 /**
  * Flags for the state of restart.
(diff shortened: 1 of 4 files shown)


lib/tcpip/tcpip-internal.h
```diff
@@ -97,31 +100,38 @@ extern std::atomic<uint8_t> userThreadCount;
...
- * manipulates `userThreadCount` manually.
+ * appropriately and that the thread is running under the error handler.  Every
+ * entry point of the TCP/IP stack API (with the exception of the driver
+ * thread, see below) should go through this unless it manipulates
+ * `userThreadCount` and sets up the error handler manually.
  */
 auto with_restarting_checks(auto operation, auto errorValue)
 {
-	if (restartState.load() != 0)
+	CHERIOT_DURING
```
davidchisnall, line 110: Since we're in C++, we can use `on_error` here and avoid the macros.

lib/tcpip/tcpip-internal.h
```diff
@@ -135,8 +145,18 @@ auto with_restarting_checks(auto operation, auto errorValue)
  */
 auto with_restarting_checks_driver(auto operation, auto errorValue)
 {
-	userThreadCount++;
-	auto ret = operation();
-	userThreadCount--;
-	return ret;
+	CHERIOT_DURING
```
davidchisnall, line 148: Same here.

lib/tcpip/tcpip-internal.h
```diff
@@ -135,8 +145,18 @@ auto with_restarting_checks(auto operation, auto errorValue)
...
+	{
+		userThreadCount++;
+		auto ret = operation();
+		userThreadCount--;
```
davidchisnall, line 152: Can this have the same comment as above?

lib/tcpip/tcpip_error_handler.h
```diff
@@ -31,24 +31,20 @@ extern struct FlagLockState ipThreadLockState;
...
  * Other locks and futexes to release as part of the reset. These are
  * documented in `sdk/include/FreeRTOS-Compat/task.h` in the main tree
- * (`__CriticalSectionFlagLock` and `__SuspendFlagLock`) and
- * `source/FreeRTOS_IP.c` in the FreeRTOS+TCP tree (`xNetworkEventQueue`).
+ * (`__CriticalSectionFlagLock` and `__SuspendFlagLock`),
+ * `source/FreeRTOS_IP.c` in the FreeRTOS+TCP tree (`xNetworkEventQueue`), and
+ * `network_wrapper.cc` in the TCP/IP wrapper.
  */
 extern struct RecursiveMutexState __CriticalSectionFlagLock;
 extern struct RecursiveMutexState __SuspendFlagLock;
```
davidchisnall, line 39: Don't these get declared in FreeRTOS.h?

lib/tcpip/tcpip_error_handler.h
```diff
@@ -31,24 +31,20 @@ extern struct FlagLockState ipThreadLockState;
...
 extern QueueHandle_t              xNetworkEventQueue;
+extern FlagLockPriorityInherited  sealedSocketsListLock;
```
davidchisnall, line 41: Missing doc comment.

davidchisnall (approved):

lib/tcpip/tcpip_error_handler.h
hlef: Actually these four locks are documented with the one big doc comment above them, which I updated accordingly in the PR?

lib/tcpip/tcpip_error_handler.h
hlef: I think that the doc comment is correct, they are documented in https://github.com/CHERIoT-Platform/cheriot-rtos/blob/main/sdk/include/FreeRTOS-Compat/task.h#L85

Or am I missing something?

hlef:
Addressed the comments. Planning to merge once I re-tested the PR later today (no access to the FPGA right now).

PR #49, opened 2024-11-01
Firewall: Send a read-only/non-capturable buffer to the TCP/IP stack.

hlef:
The firewall currently sends a writable and capturable frame buffer capability to the TCP/IP stack. This is bad because the TCP/IP stack can keep the capability and alter the buffer at a later point when we re-use it. Not sure what the exact impact is, but it sounds like the TCP/IP may be able to use this to add endpoints to the firewall table.


davidchisnall (approved):
Ooops, I thought we were doing that already.

PR #53, opened 2024-11-22
Add MQTT support for publishing retained messages

PhilDay-CT:
Adds the option to make a published message retained by the broker.

Also sets the bounds on the topic and message passed to the publishCallback so that callback can't access any other parts of the buffer.

### DIFF
--- a/include/mqtt.h
+++ b/include/mqtt.h
@@ -137,6 +137,14 @@ int __cheri_compartment("MQTT")
  *
  * `qos` indicates the level of QoS (0, 1, or 2).
  *
+ * `retain` indicates whether the message should be published as retained or
+ * not. The broker stores the last retained message and the corresponding QoS
+ * for that topic. Each client that subscribes to a topic pattern that matches
+ * the topic of the retained message receives the retained message immediately
+ * after they subscribe. The broker stores only one retained message per topic.
+ * Retained messages are cleared by publishing a zero length message with the
+ * retain flag set.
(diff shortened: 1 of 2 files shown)


nwf (approved):
LGTM

include/mqtt.h
```diff
@@ -170,7 +170,8 @@ int __cheri_compartment("MQTT") mqtt_publish(Timeout    *t,
                                              const char *topic,
                                              size_t      topicLength,
                                              const void *payload,
-                                             size_t      payloadLength);
+                                             size_t      payloadLength,
+                                             bool        retain = false);
```
hlef, line 174: Would it be possible to update the documentation of `mqtt_publish` to mention this new `retain` parameter?

include/mqtt.h
```diff
@@ -181,6 +182,14 @@ int __cheri_compartment("MQTT") mqtt_publish(Timeout    *t,
  *
  * `qos` indicates the level of QoS (0, 1, or 2).
  *
+ * `retain` indicates whether the message should be published as retained or
```
hlef, line 185: I suspect that this should be added to the documentation of `mqtt_publish`, not `mqtt_subscribe`?

PhilDay-CT: Doh !

hlef:
Looks good to me! @davidchisnall merging since Wes ACK-ed too.

PR #63, opened 2025-02-12
mqtt: bound PUBLISH callback payload inexactly

nwf:
These payloads may be larger than we can represent exactly with byte granularity.  Adjacent data is merely part of the same MQTT packet in the same client's (reused) network buffer, so this is not much of an information disclosure risk.


davidchisnall:
Is it worth explicitly zeroing anything in the inexact range? If the client of this API is a multiplexer, I guess it could do that if the length argument of the callback doesn’t match the length? Might be worth documenting that, at least.

nwf:
Hm.  Refining my original assertion, I _think_, from looking at https://github.com/FreeRTOS/coreMQTT/blob/2beef04725328923e05e576b884212d53ec97af7/source/core_mqtt_serializer.c#L1404-L1494 , that any bytes between `payload`'s `base` and `address` will be from _this_ packet and shouldn't be zeroed (because it's the topic string, for example), but data between `address + payloadLength` and `top` could be from prior packets.  We should do the zeroing before handing out the (read-only, non-capturable) pointer.  I'll add that.

PR #67, opened 2025-03-16
Various code size optimizations for the network stack

hlef:
I did a code size audit on the network stack. These changes make us gain about 4KB.

Tested on the Arty, including reset.

### DIFF
--- a/lib/sntp/sntp.cc
+++ b/lib/sntp/sntp.cc
@@ -334,6 +334,10 @@ namespace
 	{
 		static FlagLockPriorityInherited lock;
 		NetworkContext                   udpContext{timeout, nullptr};
+		SntpStatus_t                     status;
+		// Use the do {...} while {false} to deduplicate
+		// error-handling.
+		do
 		{
 			if (LockGuard guard{lock, timeout})
 			{
@@ -364,7 +368,7 @@ namespace
 				SntpContext_t context;
 
 				/* Initialize context. */
-				SntpStatus_t status =
+				status =
 				  Sntp_Init(&context,
 				            timeServers,
 				            sizeof(timeServers) / sizeof(SntpServerInfo_t),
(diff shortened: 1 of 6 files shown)


lib/tcpip/tcpip_error_handler.h
```diff
@@ -12,7 +12,7 @@
 #include <simulator.h>
 #include <vector>
 
-using DebugErrorHandler = ConditionalDebug<true, "TCP/IP Stack error handler">;
+using DebugErrorHandler = ConditionalDebug<false, "TCP/IP Stack error handler">;
```
nwf, line 15: 8012640fa6732e3979fe448505b2e68e4264accc adds a `xmake.lua` `--debug-TCPIP` option that should probably be used here.

lib/tcpip/network_wrapper.cc
```diff
@@ -460,8 +460,9 @@ Socket network_socket_create_and_bind(Timeout            *timeout,
                                       bool                isListening,
                                       uint16_t            maxConnections)
 {
-	return with_restarting_checks(
-	  [&]() -> Socket {
+	Socket ret = nullptr;
+	with_restarting_checks(
```
nwf, line 464: Is the return of this function lost?

lib/sntp/sntp.cc
```diff
@@ -334,6 +334,10 @@ namespace
 	{
 		static FlagLockPriorityInherited lock;
 		NetworkContext                   udpContext{timeout, nullptr};
+		SntpStatus_t                     status;
+		// Use the do {...} while {false} to deduplicate
+		// error-handling.
```
nwf, line 339: IMHO, don't be afraid to use `goto`s for this kind of control flow.  I know that's not a universal opinion.

lib/tcpip/network_wrapper.cc
hlef: Actually we don't care about that return value, the error is already encoded in the socket being null or not.

The function has to return an int to match the signature expected by `with_restarting_checks` to avoid another template instantiation.

I used the differentiation between -EAGAIN, -ENOMEM, etc. for readability, making sense visually of the error paths. But I realize now it may be confusing. Any suggestion?

lib/sntp/sntp.cc
hlef: Agree! Happy to move to a goto if you would prefer that. @davidchisnall any opinion?

hlef: ACK, should we merge this as it is now then?

PR #88, opened 2026-05-15
Add HIL CI workflow for testing network-stack examples on SONATA

wbeasley-thegoodpenguin:
Uses The Capable Hub's board farm to run the network-stack examples on a SONATA board.

Building and flashing are split into separate jobs to minimise occupation time in the board farm.

A couple of tweaks vs the documented build/run process:
- IPv6 is disabled for the tests as there are known issues.
- The ephemeral VM in the board farm has a limited DHCP range for the target, and the board's MAC changes between tests. The table is cleared between runs to ensure a clean slate.

### DIFF
--- a/.github/sonata-hil/conftest.py
+++ b/.github/sonata-hil/conftest.py
@@ -1,0 +1,8 @@
+import pytest
+
+
+@pytest.fixture(scope="session")
+def console(target):
+    serial = target.get_driver("ConsoleProtocol")
+    target.activate(serial)
+    return serial
--- a/.github/sonata-hil/run-hil-tests.sh
+++ b/.github/sonata-hil/run-hil-tests.sh
@@ -1,0 +1,55 @@
+#!/usr/bin/env bash
+set -e
+
+source /opt/labgrid-env/bin/activate
+
+declare -A EXPECTED_STRINGS
+EXPECTED_STRINGS["01.SNTP"]="Current UNIX epoch time:"
+EXPECTED_STRINGS["02.HTTP"]='<a href="https://iana.org/domains/example">Learn more</a>'
+EXPECTED_STRINGS["03.HTTPS"]="No content length header found"
+EXPECTED_STRINGS["04.MQTT"]="Done testing MQTT."
+EXPECTED_STRINGS["05.HTTP_SERVER"]="Listening on port"
+
+TESTS=("01.SNTP" "02.HTTP" "03.HTTPS" "04.MQTT" "05.HTTP_SERVER")
+
+declare -A RESULTS
+overall_result=0
+
+for TEST in "${TESTS[@]}"; do
+  EXPECTED="${EXPECTED_STRINGS[$TEST]}"
+  echo "--- Flashing $TEST firmware ---"
+  cp $(find firmware/$TEST -name "firmware.uf2") "$MOUNT_POINT/"
+  sync
+  sleep 5
+  echo "--- Resetting board ---"
+  sudo usbrelay HURTM_1=1
+  echo "--- Running $TEST test (timeout: 60s, expecting: '$EXPECTED') ---"
+  set +e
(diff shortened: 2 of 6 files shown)


.github/sonata-hil/conftest.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+
+
+@pytest.fixture(scope="session")
+def console(target):
+    serial = target.get_driver("ConsoleProtocol")
+    target.activate(serial)
+    return serial
```
davidchisnall, line 8: Can you add some comments or a README in this directory?  What is calling this and where?

.github/workflows/sonata-hil-ci.yml
```diff
@@ -0,0 +1,242 @@
...
+        echo "--- Serial output ---"
+        sudo find /tmp/labgrid-log -type f -exec cat {} \; || true
+        sudo rm -rf /tmp/labgrid-log
+        echo "--- Cleaning up ---"
+        rm -f "$MOUNT_POINT"/*.uf2
+        # Board MAC can change for different tests so wipe the dhcp lease table after each test
+        sudo systemctl stop dnsmasq
+        sudo rm -f /var/lib/misc/dnsmasq.leases
+        sudo systemctl start dnsmasq
+        exit $RESULT
```
davidchisnall: There's a lot of duplication here that it would be good to factor out to a script.

wbeasley-thegoodpenguin: Originally I just looped around the examples much like the build step and ran them, but I unrolled the loop when i thought there may be more complex tests in the future (resetting etc) happy to switch back to that in the name of simplicity.

.github/sonata-hil/conftest.py
wbeasley-thegoodpenguin: Will do

.github/sonata-hil/run-hil-tests.sh
```diff
@@ -0,0 +1,55 @@
...
+set -e
+
+source /opt/labgrid-env/bin/activate
+
+declare -A EXPECTED_STRINGS
+EXPECTED_STRINGS["01.SNTP"]="Current UNIX epoch time:"
+EXPECTED_STRINGS["02.HTTP"]='<a href="https://iana.org/domains/example">Learn more</a>'
+EXPECTED_STRINGS["03.HTTPS"]="No content length header found"
+EXPECTED_STRINGS["04.MQTT"]="Done testing MQTT."
+EXPECTED_STRINGS["05.HTTP_SERVER"]="Listening on port"
```
davidchisnall, line 11: I meant to provide this bit in the environment from the runner, but factor out the 'run this firmware and see if the output looks like this' bit.

.github/workflows/sonata-hil-ci.yml
```diff
@@ -0,0 +1,78 @@
...
+          xmake f --sdk=/cheriot-tools/ --board=sonata-1.3 -m release --IPv6=false
+          xmake run
+        ); done
+
+    - name: Upload firmware artifacts
+      uses: actions/upload-artifact@v7
+      with:
+        name: sonata-firmware
+        path: examples/*/build/**/firmware.uf2
+        if-no-files-found: error
```
davidchisnall, line 46: Can this not be added as a conditional step [here](https://github.com/CHERIoT-Platform/network-stack/blob/1e92f52bed7ab96df697b68b4707d8cec11e7636/.github/workflows/main.yml#L64), rather than partially duplicating the logic?

.github/sonata-hil/run-hil-tests.sh
wbeasley-thegoodpenguin: Right so leave the parts handling running the tests in their own script, and keep the tests level bits (timeouts / expected strings at the workflow level?

Issue #91, opened 2026-06-25
Use user permissions on sockets


hlef:
Enabled by https://github.com/CHERIoT-Platform/cheriot-rtos/pull/614 and similar to what we recently did [in the allocator](https://github.com/CHERIoT-Platform/cheriot-rtos/pull/617) and [in the message queue](https://github.com/CHERIoT-Platform/cheriot-rtos/pull/618/), we can use the three currently-unused bits at the bottom of sealed socket handles to store permissions.

Having permissions on sockets would allow us to encode the right to close a socket, send, or receive data. (Or more, ideas welcome!)

We could for example use these to enable compartmentalized applications that separate receiving data and sending data. If you do that with the existing socket API (or with other compartmentalized systems), you need three compartments: one that holds the socket, one that processes incoming data, and one that processes outgoing data. With socket permissions, you can eliminate the compartment that holds the socket, and just have the compartment that processes incoming data (holding a receive-only socket), and one that send data (holding a send-only socket). It reduces the number of domain switches on the hot path.

PR #94, opened 2026-07-02
Use user permissions on socket handles

hlef:
This addresses #91. While at it, it adds some missing documentation (which I realized while documenting the feature), and adds a missing check (which I found out while adding the permission checks).

This comes with a new example to showcase the use of least-privilege socket handles.

### DIFF
--- a/examples/06.COMPARTMENTALIZED_HTTP_SERVER/http_server.cc
+++ b/examples/06.COMPARTMENTALIZED_HTTP_SERVER/http_server.cc
@@ -1,0 +1,251 @@
+#include "http_server.h"
+#include "timeout.h"
+#include <NetAPI.h>
+#if __has_include(<allocator.h>)
+#	include <allocator.h>
+#endif
+#include <debug.hh>
+#include <fail-simulator-on-error.h>
+#include <thread.h>
+#include <tick_macros.h>
+
+using CHERI::Capability;
+
+using Debug            = ConditionalDebug<true, "HTTP server example test">;
+constexpr bool UseIPv6 = CHERIOT_RTOS_OPTION_IPv6;
+
+/**
+ * Bind capability for the server port 80. Use IPv6 if enabled in the
+ * configuration, and allow at most two simultaneous connections to the server.
+ */
+#define LISTEN_PORT 80
+DECLARE_AND_DEFINE_BIND_CAPABILITY(HTTPPort, UseIPv6, LISTEN_PORT, 10);
+
+DECLARE_AND_DEFINE_ALLOCATOR_CAPABILITY(TestMalloc, 32 * 1024);
+#define TEST_MALLOC STATIC_SEALED_VALUE(TestMalloc)
+
+/**
+ * Maximum number of clients the server will server before shutting down. This
+ * is useful to check that the server can handle multiple clients before
+ * terminating.
+ */
+static const uint16_t MaxClients = 10;
+
+/**
+ * This example server is written to showcase the network stack server API, but
+ * also the network stack restart. If one of the network stack APIs fails in a
+ * way likely caused by a crash, the server will wait this small delay before
+ * retrying to give time to the network stack to reset.
+ */
+static const uint16_t RestartDelay = 100; // in ticks
+
+/**
+ * Helper to call after a presumed network stack crash to wait a little bit
+ * until it completed a reset.
+ */
+inline void sleep_after_crash()
+{
+	Timeout sleep{RestartDelay};
+	thread_sleep(&sleep);
+};
+
+/**
+ * Helper to close a socket with support for network stack crashes.
+ *
+ * The close() API returns immediately with -EAGAIN if the network stack is
+ * currently resetting, hence the need for the loop. Ideally users would not
+ * have to write that helper or loop - this should be fixed in the network
+ * stack API.
+ */
+inline bool network_socket_close_retry(auto socket)
+{
+	Timeout unlimited{UnlimitedTimeout};
+	int     retries = 10;
+	for (; retries > 0; retries--)
+	{
+		int ret = network_socket_close(&unlimited, TEST_MALLOC, socket);
+		if (ret == 0)
+		{
+			return true;
+		}
+		else if (ret != -EAGAIN)
+		{
+			return false;
+		}
+		// The network stack is still resetting.
+		sleep_after_crash();
+	}
+	return false;
+};
+
+void debug_network_address(uintptr_t value, DebugWriter &writer)
+{
+	auto *address = reinterpret_cast<NetworkAddress *>(value);
+	if (address->kind == NetworkAddress::AddressKindIPv6)
+	{
+		for (int i = 0; i < 14; i += 2)
+		{
+			writer.write_hex_byte(address->ipv6[i]);
+			writer.write_hex_byte(address->ipv6[i + 1]);
+			writer.write(':');
+		}
+		writer.write_hex_byte(address->ipv6[14]);
+		writer.write_hex_byte(address->ipv6[15]);
+	}
+	else if (address->kind == NetworkAddress::AddressKindIPv4)
+	{
+		writer.write_decimal((address->ipv4 >> 0) & 0xff);
+		writer.write('.');
+		writer.write_decimal((address->ipv4 >> 8) & 0xff);
+		writer.write('.');
+		writer.write_decimal((address->ipv4 >> 16) & 0xff);
+		writer.write('.');
+		writer.write_decimal((address->ipv4 >> 24) & 0xff);
+	}
+	else
+	{
+		writer.write("<invalid address>");
+	}
+};
+
+template<>
+struct DebugFormatArgumentAdaptor<NetworkAddress>
+{
+	static DebugFormatArgument construct(NetworkAddress &address)
+	{
+		return {reinterpret_cast<uintptr_t>(&address),
+		        reinterpret_cast<uintptr_t>(debug_network_address)};
+	}
+};
+
+void __cheri_compartment("compartmentalized_http_server_example") example()
+{
+	// TODO We can eliminate this object once network APIs have been ported
+	// to the new timeout API.
+	Timeout unlimited{UnlimitedTimeout};
+
+	network_start();
+
+	auto heapAtStart = heap_quota_remaining(TEST_MALLOC);
+
+	Debug::log("Starting the server.");
+
+	// If the network stack crashes, the listening socket will be closed
+	// (as will all accepted sockets). The outer loop recreates the
+	// listening socket if this happens, up until we have handled the
+	// correct number of clients.
+	uint16_t clientsCounter = 0;
+	while (clientsCounter < MaxClients)
+	{
+		Debug::log("Creating a listening socket.");
+		auto socket = network_socket_listen_tcp(
+		  &unlimited, TEST_MALLOC, STATIC_SEALED_VALUE(HTTPPort));
+
+		if (!Capability{socket}.is_valid())
+		{
+			Debug::log("Failed to create a listening socket.");
+			// This may have failed because of a network stack
+			// crash. Sleep a little bit to enable a reset.
+			sleep_after_crash();
+			continue;
+		}
+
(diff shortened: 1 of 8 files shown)


examples/06.COMPARTMENTALIZED_HTTP_SERVER/http_server.cc
```diff
@@ -0,0 +1,178 @@
...
+			// server). We could reduce this to three (server ->
+			// receiver -> sender -> server) by sharing the sockets
+			// through shared variables and having the receive
+			// compartment call the send compartment directly.
+			int received = example_receive(
+			  network_socket_permissions_and(clientSocket, SocketPermitReceive),
+			  allocator_permissions_and(TEST_MALLOC, AllocatorPermitAllocate));
+			if (received > 0)
+			{
+				example_send(network_socket_permissions_and(clientSocket,
```
davidchisnall: It would be nice to also remove Global from the socket handle. That gives you some tight guarantees: the callee can send via this socket, only for the duration of the call.

examples/06.COMPARTMENTALIZED_HTTP_SERVER/http_server.cc
```diff
@@ -0,0 +1,178 @@
...
+
+	uint16_t clientsCounter = 0;
+
+	Debug::log("Starting the server.");
+
+	while (clientsCounter < MaxClients)
+	{
+		Debug::log("Creating a listening socket.");
+		Timeout unlimited{UnlimitedTimeout};
+		auto    socket = network_socket_listen_tcp(
```
davidchisnall: I initially found this confusing.  The outer loop, as I (now) understand it, exists so that the network stack can crash and we'll recreate the listening socket and go back to listening.  That's a nice thing to demo, but it needs a comment on the outer loop, something along the lines of:

If the network stack crashes, the listening socket will be closed (as will all accepted sockets).  The outer loop recreates the listening socket if this happens, up until we have handled the correct number of clients.

examples/06.COMPARTMENTALIZED_HTTP_SERVER/http_server.cc
```diff
@@ -0,0 +1,178 @@
...
+	auto heapAtStart = heap_quota_remaining(TEST_MALLOC);
+
+	uint16_t clientsCounter = 0;
+
+	Debug::log("Starting the server.");
+
+	while (clientsCounter < MaxClients)
+	{
+		Debug::log("Creating a listening socket.");
+		Timeout unlimited{UnlimitedTimeout};
```
davidchisnall: With the new timeout work, we can replace this with `TimeoutWaitForever`.

examples/06.COMPARTMENTALIZED_HTTP_SERVER/http_server.cc
```diff
@@ -0,0 +1,178 @@
...
+		auto    socket = network_socket_listen_tcp(
+		  &unlimited, TEST_MALLOC, STATIC_SEALED_VALUE(HTTPPort));
```
davidchisnall, line 142:
```suggestion
		  TimeoutWaitForever, TEST_MALLOC, STATIC_SEALED_VALUE(HTTPPort));
```

examples/06.COMPARTMENTALIZED_HTTP_SERVER/http_server.cc
```diff
@@ -0,0 +1,178 @@
...
+		{
+			NetworkAddress clientAddress = {0};
+			uint16_t       clientPort    = 0;
+
+			auto clientSocket = network_socket_accept_tcp(
+			  &unlimited, TEST_MALLOC, socket, &clientAddress, &clientPort);
+
+			if (!Capability{clientSocket}.is_valid())
+			{
+				Debug::log("Failed to establish a connection.");
```
davidchisnall: Accept maybe?

examples/06.COMPARTMENTALIZED_HTTP_SERVER/http_server.cc
```diff
@@ -0,0 +1,178 @@
...
+				Timeout sleep{RestartDelay};
+				thread_sleep(&sleep);
```
davidchisnall: This fragment appears several times in this function, maybe factor it out into a lambda?

examples/06.COMPARTMENTALIZED_HTTP_SERVER/http_server.cc
```diff
@@ -0,0 +1,178 @@
...
+				break;
+			}
+
+			if (!UseIPv6)
+			{
+				Debug::log("Established a connection with {}.{}.{}.{}:{}",
```
davidchisnall: I could have sworn I added a debug helper to print IP addresses.  Did I not?

davidchisnall: Aha, I did it for [one of the book examples](https://github.com/CHERIoT-Platform/book-examples/blob/98b8311e9de408677be29f90c8d37b3d89307d7c/examples/debug_helpers/example.cc#L34).  We should pull it in here.

examples/06.COMPARTMENTALIZED_HTTP_SERVER/http_server.cc
```diff
@@ -0,0 +1,178 @@
...
+			// allocator capability for allocating and freeing. The
+			// send compartment only needs a socket for sending.
+			//
+			// This implementation requires four compartment
+			// switches (server -> receiver -> server -> sender ->
+			// server). We could reduce this to three (server ->
+			// receiver -> sender -> server) by sharing the sockets
+			// through shared variables and having the receive
+			// compartment call the send compartment directly.
+			int received = example_receive(
```
davidchisnall: Even though it's an example, it would be nice if this returned a parsed request, even if it's a simple very + path and in the example it's always GET and /.  Then we can show that a real system would have some map from paths to handlers and allow privilege separating paths (e.g. admin interface has the ability to do things that a read-only diagnostic interface simply can't do).

examples/06.COMPARTMENTALIZED_HTTP_SERVER/http_server.cc
```diff
@@ -0,0 +1,178 @@
...
+			  allocator_permissions_and(TEST_MALLOC, AllocatorPermitAllocate));
+			if (received > 0)
+			{
+				example_send(network_socket_permissions_and(clientSocket,
+				                                            SocketPermitSend));
+			}
+
+			Debug::log("Terminating the connection with the client.");
+			int retries = 10;
+			// In a retry loop to be more rebust to network stack crashes.
```
davidchisnall: Hmm, can close fail if the network stack crashes?  That sounds like a bug, if the stack dies in the middle we should free the socket structure and report that we succeeded.

examples/06.COMPARTMENTALIZED_HTTP_SERVER/http_server.cc
```diff
@@ -0,0 +1,178 @@
...
+				thread_sleep(&sleep);
+			}
+
+			if (retries == 0)
+			{
+				Debug::log("Failed to close the client socket.");
+			}
+		}
+
+		Debug::log("Closing the listening socket.");
```
davidchisnall, line 228: Same here.  I'm not happy making people write loops here, we should move this loop into `network_socket_close`.

examples/06.COMPARTMENTALIZED_HTTP_SERVER/http_server.h
```diff
@@ -0,0 +1,6 @@
+#include <NetAPI.h>
+
```
davidchisnall, line 2: It would be nice to have doc comments on these.

davidchisnall: It would also be nice to give them names that describe what they're supposed to do, for example `http_parse_request` and `http_send_response`.

examples/06.COMPARTMENTALIZED_HTTP_SERVER/send_compartment.cc
```diff
@@ -0,0 +1,43 @@
+#include "timeout.h"
+#include <NetAPI.h>
+#include <debug.hh>
+#include <fail-simulator-on-error.h>
+
+using Debug =
+  ConditionalDebug<true, "HTTP server example test (send compartment)">;
+
+static char reply[] =
```
davidchisnall, line 14: Is there a reason this isn't const?

examples/06.COMPARTMENTALIZED_HTTP_SERVER/http_server.cc
hlef: The APIs need to be modified for that, right? I'd maybe leave that change for later when they are modified and all examples can be changed in bulk.

examples/06.COMPARTMENTALIZED_HTTP_SERVER/http_server.cc
hlef: (See comment above on the network APIs)

examples/06.COMPARTMENTALIZED_HTTP_SERVER/http_server.cc
hlef: This would be great, but as discussed offline there seems to be a limitation in the compiler that makes it tricky to remove G on sealed caps. We'll leave a TODO for now.

examples/06.COMPARTMENTALIZED_HTTP_SERVER/http_server.cc
hlef: Right now the close() API will return immediately with `-EAGAIN` if the network stack is currently resetting, hence the need for the loop. I agree that we shouldn't ask users to write this loop!

I cannot fix this right as part of this PR but I will open a bug report, we can give it some cycles when we will work on fixing limitations of the reset.

examples/06.COMPARTMENTALIZED_HTTP_SERVER/http_server.cc
hlef: (Comment above on that)

examples/06.COMPARTMENTALIZED_HTTP_SERVER/send_compartment.cc
hlef: The network stack API takes a non-const pointer.

examples/06.COMPARTMENTALIZED_HTTP_SERVER/http_server.cc
hlef: See https://github.com/CHERIoT-Platform/network-stack/issues/98.

examples/06.COMPARTMENTALIZED_HTTP_SERVER/http_server.cc
hlef: See https://github.com/CHERIoT-Platform/llvm-project/issues/430.

examples/06.COMPARTMENTALIZED_HTTP_SERVER/http_server.cc
hlef: Just to make sure this doesn't get forgotten: https://github.com/CHERIoT-Platform/network-stack/issues/99

hlef:
Alright @davidchisnall, addressed all I could and created issues for everything else. Let me know what I broke in that process :slightly_smiling_face:

lib/tcpip/network_wrapper.cc
```diff
@@ -725,7 +736,8 @@ Socket network_socket_accept_tcp(Timeout            *timeout,
 		  socket = sealedSocket;
 		  return 0;
 	  },
-	  sealedListeningSocket);
+	  sealedListeningSocket,
+	  0 /* No specific permission required */);
```
davidchisnall, line 740: I wonder if we should require read here?

davidchisnall:
Looks good.  We don't normally use caps in example names (the existing ones are just initialisms, which *are* capitalised, but the rest shouldn't be).

Issue #104, opened 2026-07-18
Race condition in network_socket_close in reset scenario


nding0405:
### Summary
If two threads call `network_socket_close()` concurrently on the same socket after a network stack restart, both may enter the cleanup path at the same time. One thread can destroy the socket wrapper while the other tries to dereference it, that will cause Use-After-Free, trigger hardware exception, and lead to another network stack restart.

### Root cause
For a socket from last epoch, both threads fail to acquire `socketLock`, but the the `network_socket_close` allows both threads to continue when it finds this is a restart case. The race can occur near [heap_free()](https://github.com/CHERIoT-Platform/network-stack/blob/075337c1197aec5abf9078684b3f901382f198ce/lib/tcpip/network_wrapper.cc#L992), and here is the problematic order of interleaving.

```text
Thread 1                              Thread 2
some prior checks
                                      some prior checks
heap_free(..., socket->socket)
token_obj_destroy(..., sealedSocket)
                                      *heap_free(..., socket->socket) // crash
```
After Thread 1 destroys the socket wrapper, Thread 2 dereferences the freed wrapper to read `socket->socket`. This triggers a harware fault and thus lead to another network stack restart.

### Affected component
`lib/tcpip/network_wrapper.cc`, old epoch cleanup path in `network_socket_close()`.

### Impact
A caller that closes the same old socket concurrently from two threads stand a chance to crash and restart the network stack again during reset, which is not expected. But this is not a fatal error, the bug won't be triggered after the second crash, since the socket wrapper has already been freed after the first crash, `with_sealed_socket` will detect that and return the error code instead of triggering another hardware exception.

### PoC
Here is the test code. The test adds a guard immediately before [`heap_free()`](https://github.com/CHERIoT-Platform/network-stack/blob/075337c1197aec5abf9078684b3f901382f198ce/lib/tcpip/network_wrapper.cc#L992) to force the race. Both threads stop at the guard. The fast thread will continue first and executes `heap_free()` and `token_obj_destroy()` together. The delayed thread is then released and dereferences the freed socket wrapper, triggering the hardware exception.
```c
poc_test.cc
void __cheri_compartment("http_server_example") example()
{
	// ... some initializations
	Debug::log("Injecting network-stack fault.");
	network_inject_fault();
	Timeout sleep{1000};
        /**
         * During that time gap, send a packet to
         * the network stack to trigger the reset.
         */
	thread_sleep(&sleep); 

	thread_pool::async([socket]() {
[...]

PR #106, opened 2026-07-22
Fixed with_sealed_socket UAF issue

nding0405:
This PR partially addresses issue #104. It adds an ephemeral claim in `with_sealed_socket()` to prevent a Use-After-Free when the socket is dereferenced before acquiring its lock. The remaining case mentioned in #104 that needs to be fixed is in `network_socket_close()`, where we need another ephemeral claim to prevent a Use-After-Free if the lock has already been destroyed during a reset.

### DIFF
--- a/lib/tcpip/network_wrapper.cc
+++ b/lib/tcpip/network_wrapper.cc
@@ -120,7 +120,7 @@ namespace
 			  }
 			  /*
 			   * We are not holding the socket lock here, but still doing
-			   * deference of the socket, which can lead to an UAF error.
+			   * dereference of the socket, which can lead to an UAF error.
 			   * To address this, we use ephemeral call before dereferencing
 			   * the socket.
 			   */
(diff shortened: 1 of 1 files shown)


lib/tcpip/network_wrapper.cc
```diff
@@ -118,6 +118,23 @@ namespace
...
 				  return -EINVAL;
 			  }
+			  Timeout unlimited{UnlimitedTimeout};
+			  /**
+			   * We are not holding the socket lock here, but still doing
+			   * deference of the socket, which can lead to an UAF error.
+			   * To address this, we use ephemeral call before dereferencing
+			   * the socket.
+			   */
+			  int result = heap_claim_ephemeral(&unlimited, socket, nullptr);
```
davidchisnall: The idiomatic way of writing this is:

```c++
int result = heap_claim_ephemeral(TimeoutWaitForever, socket, nullptr);
```

That said, I'm not entirely comfortable having a thing with an unlimited timeout inside APIs that all take a timeout.  We should be propagating that down.

nding0405: Shall we change the signature of the lockless `with_sealed_socket`? It currently doesn't take the timeout argument.

davidchisnall (approved):

PR #107, opened 2026-07-23
TCPIP: make the return value of network_socket_accept_tcp more expressive

nding0405:
The network_socket_accept_tcp function can only return either a valid sealed capability or an untagged nullptr, so the caller cannot distinguish a timeout from an out-of-memory failure. Change it so that it can return an invalid capability that encodes the error code.

### DIFF
--- a/include/NetAPI.h
+++ b/include/NetAPI.h
@@ -235,8 +235,14 @@ Socket __cheri_compartment("NetAPI")
  * of the connected client.  These can be null if the caller is not interested
  * in the client's address or port.
  *
- * This returns a valid sealed capability to a connected socket on success, or
- * an untagged value on failure.
+ * This returns either a valid sealed socket type or an untagged capability that
+ * encodes the error code, and here are what each of them represent:
+ * - `-ENOMEM`: allocation of the wrapper timed out or does not have memory for
+ * now.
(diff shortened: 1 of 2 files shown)


include/NetAPI.h
```diff
@@ -235,8 +235,13 @@ Socket __cheri_compartment("NetAPI")
...
  *
- * This returns a valid sealed capability to a connected socket on success, or
- * an untagged value on failure.
+ * This returns either a valid sealed socket type or an untagged capability that
+ * encodes the error code, and here are what each of them represent:
+ * -ENOMEM: allocation of the wrapper timed out or does not have memory for now.
+ * -EINVAL: The timeout pointer is invalid, or FreeRTOS_accept returns invalid
+ * 			socket, or fail to claim the socket poiting to mallocCapability,
+ * 			or fail to add socket to the socket reset list,
+ * -ETIMEDOUT: timed out on FreeRTOS_accept.
```
davidchisnall: Please can you put backticks around the values (e.g. `-ENOMEM`) and make them a Markdown list?

davidchisnall (approved):
Looks good, minor comment on a comment.

include/NetAPI.h
nding0405: Hi David, since #107 has been approved. Would you mind merging it when you have a chance? Once it is merged I will rebase #109 and re-request review. Thanks!

PR #109, opened 2026-07-26
Add multi-waiter support for TCP accept

nding0405:
### New feature description:
Allow a single thread to wait for incoming connections across multiple
listening sockets at once using multiwaiter_wait(), instead of having to
block in network_socket_accept_tcp() on one socket at a time.

Each socket now exposes an array of futex called eventFutexState, that 
a caller can register them with a multiwaiter. For this PR, only the added
the accept functionality, others remains commented out for now. A thread 
can watch the accept futexes of several listening sockets simultaneously, 
block in multiwaiter_wait() until one of them signals a ready connection, and then 
accept that connection. The blocking wait therefore moves out of accept() and 
into the multiwaiter: accept() is called only once a connection is known to be 
ready, so it no longer has to block, which is what makes it possible to wait on 
several listening sockets at the same time from a single thread.

### Key changes:
1. Added Event futex array (tcpip-internal.h, NetAPI.h) for socket wrapper: Only
SocketAcceptEvent is implemented today, room reserved for future receive/send 
events. The value of the futex counts the ready events currently pending on the
socket. For SocketAcceptEvent, it represents the number of connections that are 
ready to be accepted. The reserved value SocketNotAvailable (INT_MIN) means the
socket will be freed soon.

2. Read-only getter: network_socket_get_event_source() returns a capability to
a socket's event futex stripped to read-only.

3. Producing events: on_tcp_connect increments the SocketAcceptEvent futex and calls
notify_all() when a new connection becomes ready. To let the callback can find
the wrapper from the raw socket, network_socket_create_and_bind() now applys 
a bidirectional link between the wrapper and raw socket.

4. Consuming events: network_socket_accept_tcp() decrements the futex after it
successfully accepts a connection. This only keeps the futex value align with the 
semantic desbribed above. It does not call notify_all(), since dequeuing an existing 
connection is not a new, so no need to wake the threads up.

5. Avoid waiting forever when socket is not available: the futex lives in the socket
wrapper's heap memory, so a thread must never be left blocked on a socket
that is being freed. Both network_socket_close() and the reset handler
(reset_network_stack_state) set every event futex to SocketNotAvailable and
call notify_all() before that memory goes away. The value change wakes any
thread blocked in multiwaiter_wait(), which then observes the sentinel and
returns instead of sleeping on a dead socket. A subsequent accept() will see the
sentinel and reports the socket as unavailable.

### DIFF
--- a/include/NetAPI.h
+++ b/include/NetAPI.h
@@ -40,8 +40,12 @@ struct NetworkAddress
  */
 enum SocketEventType : uint8_t
 {
-	SocketAcceptEvent = 0, // Triggered when a new TCP connection is accepted on
-	                       // a listening socket
+	/// Triggered when a new TCP connection
+	/// is accepted on a listening socket.
+	/// The value of the futex corresponding to
+	/// SocketAcceptEvent, means the number of
+	/// pending connections.
+	SocketAcceptEvent = 0,
 	// SocketReceiveEvent = 1, // Triggered when data is received on a socket
 	// SocketSendEvent = 2     // Triggered when a socket has space available
 	// for sending
--- a/include/NetAPI.h
+++ b/include/NetAPI.h
@@ -35,6 +35,25 @@ struct NetworkAddress
 };
 
 /**
+ * Enumeration that defines futex types. Each value corresponds
+ * to an index in the socket's futex array.
+ */
+enum SocketEventType : uint8_t
+{
+	SocketAcceptEvent = 0, // Triggered when a new TCP connection is accepted on
+	                       // a listening socket
+	// SocketReceiveEvent = 1, // Triggered when data is received on a socket
+	// SocketSendEvent = 2     // Triggered when a socket has space available
+	// for sending
+};
+
+/// Number of distinct socket event futex types.
+static constexpr size_t NumFutexTypes = SocketEventType::SocketAcceptEvent + 1;
+/// Sentinel value stored in a socket event futex after the socket has been
+/// torn down.
+static constexpr uint32_t SocketNotAvailable = -1;
+
+/**
  * Enumeration defining the connection type.
  */
 enum ConnectionType : uint8_t
@@ -268,6 +287,14 @@ Socket __cheri_compartment("TCPIP")
   network_socket_udp(Timeout            *timeout,
                      AllocatorCapability mallocCapability,
                      bool                isIPv6);
+
+/**
+ * Return the event source associated with a socket.
+ *
+ * The returned capability is read-only and bounded to four bytes.
+ */
+uint32_t *__cheri_compartment("TCPIP")
+  network_socket_get_event_source(Socket sealedSocket, SocketEventType type);
 
 /**
  * Authorise a UDP socket to send packets to a specific host.  This opens a
--- a/include/NetAPI.h
+++ b/include/NetAPI.h
@@ -51,7 +51,7 @@ enum SocketEventType : uint8_t
 static constexpr size_t NumFutexTypes = SocketEventType::SocketAcceptEvent + 1;
 /// Sentinel value stored in a socket event futex after the socket has been
 /// torn down.
-static constexpr uint32_t SocketNotAvailable = -1;
+static constexpr int32_t SocketNotAvailable = INT32_MIN;
 
 /**
  * Enumeration defining the connection type.
--- a/lib/tcpip/network_wrapper.cc
+++ b/lib/tcpip/network_wrapper.cc
@@ -715,6 +715,13 @@ Socket network_socket_accept_tcp(Timeout            *timeout,
 		  }
 
 		  socketWrapper->socketEpoch = currentSocketEpoch.load();
+		  /*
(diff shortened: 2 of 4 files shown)


include/NetAPI.h
```diff
@@ -34,6 +34,24 @@ struct NetworkAddress
...
+	                       // a listening socket
+	// SocketReceiveEvent = 1, // Triggered when data is received on a socket
+	// SocketSendEvent = 2     // Triggered when a socket has space available
+	// for sending
+};
+
+static constexpr size_t NumFutexTypes = SocketEventType::SocketAcceptEvent + 1;
+// Sentinel stored in the socket's event futexes to mark the socket as torn
+// down.
+static constexpr uint32_t SocketNotAvailable = -1;
```
davidchisnall: Please can you add doc comments for these?

include/NetAPI.h
```diff
@@ -235,8 +253,13 @@ Socket __cheri_compartment("NetAPI")
  * of the connected client.  These can be null if the caller is not interested
  * in the client's address or port.
  *
- * This returns a valid sealed capability to a connected socket on success, or
- * an untagged value on failure.
+ * This returns either a valid sealed socket type or an untagged capability that
+ * encodes the error code, and here are what each of them represent:
+ * -ENOMEM: allocation of the wrapper timed out or does not have memory for now.
```
davidchisnall:
```suggestion
 * `-ENOMEM`: allocation of the wrapper timed out or does not have memory for now.
```

(And the same markup below)

lib/tcpip/network_wrapper.cc
```diff
@@ -451,6 +504,34 @@ static void on_tcp_connect(Socket_t socket, BaseType_t isConnected)
...
+		// threads sleeping on the corresponding futex.
+		// Use the FreeRTOS getter to get the wrapper capability we stored in
+		// it.
+		auto *wrapper =
+		  static_cast<SealedSocket *>(pvSocketGetSocketID(socket));
+		if (wrapper != nullptr)
+		{
+			/*
+			 * An ephemeral call is needed here to prevent UAF when
+			 * dereferencing the socket wrapper.
```
davidchisnall: I don't understand this.  The claim is dropped when you call `signal_event_futex` and it looks as if the only thing that's happening in the caller is pointer arithmetic, not dereference.

It's generally bad style to assume an ephemeral claim will be valid on the way into a function.  Looking inside `signal_event_futex`, it appears that it does a load and a CAS first, which could trap, and then a notify (which drops the claim).  I think moving the claim into that function would make ownership clearer.

lib/tcpip/network_wrapper.cc
```diff
@@ -720,6 +848,15 @@ Socket network_socket_accept_tcp(Timeout            *timeout,
 		  return 0;
 	  },
 	  sealedListeningSocket);
+
+	if (retVal != 0)
+	{
+		__clang_ignored_warning_push("-Wcheri-capability-misuse");
+		auto errCode = reinterpret_cast<Socket>(retVal);
+		__clang_ignored_warning_pop();
```
davidchisnall, line 859: It would be nice to have C and C++ helpers for this (separate PR).

davidchisnall (approved):

davidchisnall:
Accidentally hit approve.  There are a couple of small things that are worth fixing in this before it lands.

lib/tcpip/network_wrapper.cc
nding0405: Thanks! I see the distinction now. I’ll move the ephemeral claim into `signal_event_futex` instead, and `notify_all` will clear the it.

lib/tcpip/network_wrapper.cc
```diff
@@ -395,6 +395,68 @@ namespace
...
+	 * Returns 0 on success, or -EINVAL if the futex has been invalidated.
+	 */
+	int signal_event_futex(cheriot::atomic<uint32_t> &futex)
+	{
+		/*
+		 * An ephemeral call is needed here to prevent UAF on
+		 * futex.load();
+		 */
+		Timeout timeout{UnlimitedTimeout};
+		if (heap_claim_ephemeral(&timeout, &futex) != 0)
```
davidchisnall:
```suggestion
		if (heap_claim_ephemeral(TimeoutWaitForever, &futex) != 0)
```

But I don't really like having a thing that can block forever here.  It's *probably* fine, because losing the wake is bad, but it makes me a bit nervous.

lib/tcpip/network_wrapper.cc
```diff
@@ -395,6 +395,68 @@ namespace
...
+				futex.notify_all();
+				return 0;
+			}
+		}
+
+		return -EINVAL;
+	}
+
+	/**
+	 * Consume one pending event by decrementing the futex.
```
davidchisnall: Please can you document why this one *doesn't* need to claim the argument?  The caller will hold the lock that prevents the deallocation of the socket?

Is there a reason not to make these two methods on the socket structure?  That would make the ownership clearer.

PR #111, opened 2026-08-07
Firewall: keep the TCP hole open for the final ACK

nding0405:
During an active TCP close, FreeRTOS calls `on_tcp_connect` before sending the final ACK. At that point, the callback removed the TCP firewall hole, so the firewall dropped the ACK and the peer could not finish closing. Waiting for the TCP state change is not enough because the state changes before the ACK passes through egress.

Keep the firewall hole open until the final ACK passes through egress. `on_tcp_connect` marks the hole as closing, egress marks it as safe to remove, and `network_socket_close()` removes it afterward.

Key changes:
- Add a state field to each TCP firewall hole: `InUse` for an open connection, `InTermination` while waiting for the final ACK, and `CanBeRemoved` after that ACK passes egress. `NotFound` means that no matching hole exists.
- Change the `SmallTableBase` helpers to compare only each table's key fields. The endpoint key remains the address and ports, so existing helper calls keep their old behavior and ignore the new state field. Lookup remains `O(log n)`; only the close helpers inspect the state.
- Change the hole from `InUse` to `InTermination` in `on_tcp_connect`.
- Change it from `InTermination` to `CanBeRemoved` after the next matching packet passes egress.
- Make `network_socket_close()` wait for `CanBeRemoved` before removing the hole. A successful return guarantees that the hole was removed.

Note:
The polling [here](https://github.com/CHERIoT-Platform/network-stack/blob/58425e6a1c65c4357b037aa817ba492e7a8fa70b/lib/tcpip/network_wrapper.cc#L865) can be removed for ipv4 path, but I keep it here for ipv6. Since the firewall endpoint delayed removal and polling for firewall state logics are currently only applied to ipv4 path. The polling here still make sure we are at a safe point to free the socket, but it is not a signal of we can safely remove the firewall endpoint.

### DIFF
--- a/lib/firewall/firewall.hh
+++ b/lib/firewall/firewall.hh
@@ -114,18 +114,17 @@ void __cheri_compartment("Firewall")
 
 /**
  * State of an IPv4 TCP firewall hole during close.
- *
- *  - `InUse`: The connection is open and packets may pass.
- *  - `InTermination`: The final packet has not passed egress yet.
- *  - `CanBeRemoved`: Egress passed the final packet; the hole may be removed.
- *  - `NotFound`: No hole matches the address and ports.
  */
 enum class TCPFirewallState : uint8_t
 {
-	InUse         = 0,
+	/// The socket is active and packets may pass through this hole.
+	InUse = 0,
+	/// The socket is closing and its final ACK has not passed egress.
 	InTermination = 1,
-	CanBeRemoved  = 2,
-	NotFound      = 3,
+	/// The final ACK passed egress and the hole may be removed.
+	CanBeRemoved = 2,
+	/// No hole matches the remote address and local and remote ports.
+	NotFound = 3,
 };
 
 /**
--- a/lib/firewall/firewall.hh
+++ b/lib/firewall/firewall.hh
@@ -120,7 +120,7 @@ void __cheri_compartment("Firewall")
  *  - `CanBeRemoved`: Egress passed the final packet; the hole may be removed.
  *  - `NotFound`: No hole matches the address and ports.
  */
-enum class TCPFirewallState : uint32_t
+enum class TCPFirewallState : uint8_t
 {
 	InUse         = 0,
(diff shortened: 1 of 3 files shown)


lib/firewall/firewall.hh
```diff
@@ -112,6 +112,46 @@ void __cheri_compartment("Firewall")
...
 
+/**
+ * State of an IPv4 TCP firewall hole during close.
+ *
+ *  - `InUse`: The connection is open and packets may pass.
+ *  - `InTermination`: The final packet has not passed egress yet.
+ *  - `CanBeRemoved`: Egress passed the final packet; the hole may be removed.
+ *  - `NotFound`: No hole matches the address and ports.
+ */
+enum class TCPFirewallState : uint32_t
```
davidchisnall: Can we make the base type smaller than this (e.g. uint8_t?) so that it fits better in the small table?

lib/firewall/firewall.hh
```diff
@@ -112,6 +112,46 @@ void __cheri_compartment("Firewall")
...
+int __cheri_compartment("Firewall")
+  firewall_mark_tcpipv4_endpoint_in_termination(uint32_t remoteAddress,
+                                                uint16_t localPort,
+                                                uint16_t remotePort);
+
+/**
+ * Get a TCP hole's state without changing it.
+ *
+ * Returns the state, or `TCPFirewallState::NotFound` if no hole matches.
+ */
```
davidchisnall, line 153: Note that this can also fail for normal reasons so the return value might be a negative errno value.

davidchisnall:
Why the change to remove the comments on the values? This makes IDE integration worse.

nding0405:
> Why the change to remove the comments on the values? This makes IDE integration worse.

Fixed. I restored the per-value comments. I had moved them out since I thought it might look a bit cleaner, but I see why per-value is better. Could you please take another look? Thanks!

PR #120, opened 2026-08-17
Make ephemeral claims on topic and payload in mqtt_publish.

resistor:
This missing claim was detected by the CheriotHeapChecker static analysis.


resistor:
Actually, investigating if this might be a false positive

PR #121, opened 2026-08-18
Port the network stack to use the randombytes RTOS API rather than having each compartment re-implement rand().

resistor:


lib/dns/dns.cc
```diff
@@ -1080,7 +1071,10 @@ __cheri_compartment("DNS") int network_host_resolve(Timeout        *timeout,
 
 	// Prepare the query answer buffer and ID for the new query.
 	memset(&queryResult, 0, sizeof(NetworkAddress));
-	queryID = rand();
+	if (!randombytes(queryID))
```
davidchisnall: Our style recommends using explicit != 0, rather than relying on implicit conversion to bool, to make it clear that the return is not Boolean.

It might be better to return the error code that randombytes returns, rather than -EAGAIN. For example, if this returns a not-enough-stack / trusted stack error then the caller can’t just retry and expect success in the future.

davidchisnall (commented):
I left one comment but it applies across the rest.

I’m not sure when we can merge this. We try hard to avoid the requirement to version bump the RTOS and network stack at the same time, but these all use the helper that was added to the RTOS yesterday, so merging this means anyone using this needs to have yesterday’s RTOS or newer. The helper means that we can’t just gate this on the existence of the randombytes.h header and fall back to the old version.

Once we merge this, we’ll have a flag day. We should delete all of the old compatibility code in other places at the same time.

lib/dns/dns.cc
resistor: Done

resistor:
> Once we merge this, we’ll have a flag day. We should delete all of the old compatibility code in other places at the same time.

Actually, I came up with some concepts trickery to allow this network stack to re-define the wrappers if they're not available from the RTOS. That should let us avoid a flag day.

davidchisnall (approved):

